PR #160, opened 2018-08-31
Add initial support for Python 3 and Windows

ryanfitzsimon:
I don't believe this project has any fundamental incompatibilities with Python 3 or Windows.

This pull request adds initial support for both.

When testing with Python 3 I've noticed some problems with file encodings other than the system default, but this is still a big improvement over not being able to run at all.


ryanfitzsimon:
The Codacy and Code Climate checks look like they're failing due to issues that were already present in bin/vsg, but weren't being checked.

Should these be fixed in this pull request or tackled separately?

jeremiah-c-leary:
Hey Ryan,

I appreciate the two pull requests.  I will admit I'm a beginner with GitHub, but I will pull in your changes.  I just found another issue that maybe you have resolved and will be working it tonight.

Again...thanks for the updates.

--Jeremy

PR #254, opened 2019-09-24
Add rule for checking if generic name starts with G_.

m-kru:


jeremiah-c-leary:
Hey m-kru,
Nice check.  I will get it incorporated.

PR #262, opened 2019-10-02
Case rule concept [Do not merge]

m-kru:
Here are the changes that introduce `case_rule` concept. I tried to apply this new `case_rule` to 2 constant rules: `rule_002` and `rule_004`.
If you run tests, there is an issue with fixing violations. How violations are fixed now is not optimal for the new `case_rule`. I can't figure out why the test fails, I would appreciate if you take a look.

Please let me know what you think, as I do not want to move forward with this idea, if you think it is bad and is not going to be accepted.

**EDIT:** I have fixed test issue and improved a little bit how words are fixed. Only words that need fixing are fixed. Previously correct words like `constant` were replaced by correct word `constant`.


jeremiah-c-leary:
Hey m-kru,
I just saw this and will take a look at it.

jeremiah-c-leary:
I think I see what you are doing.
  1)  combined the individual functions that extracted out the signal, constant etc.. names
  2)  created another function to extract out the class name:  signal, constant, etc...
  3)  created a new test class case_rule that you pass either 1 or 2 as an argument when you create a rule to test for a specific lower/uppercase instance.

That is pretty slick.  Definitely more advanced python than I have written.

I like it, I think you should continue.

m-kru:
@jeremiah-c-leary 
It is finished from functional point of view. Let me know what you think!
I have changed my mind and used abstract method `_extract` in `case_rule`.
The docs need to be updated. I also want to improve few things in the sources, and add more tests.

jeremiah-c-leary:
I just pulled your update and used this configuration:

``` yaml
rule:
  global:
    case : 'upper'
```

and it uppercased a lot of keywords.  That was really slick.

That was a great idea you had.

Thank you.

m-kru:
Docs are still not updated!

PR #264, opened 2019-10-12
Update docs.

m-kru:
While updating the docs I have realized, that there are still some case rules that use fixed lowercase, they should also be refactored to used `case_rule`. What is more, the docs should probably also inform user, how to change case from default to lower or upper.

### DIFF
--- a/docs/attribute_rules.rst
+++ b/docs/attribute_rules.rst
@@ -33,7 +33,9 @@ This rule checks the indent of **attribute** declarations.
 attribute_002
 #############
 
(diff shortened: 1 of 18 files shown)


jeremiah-c-leary:
How about under the configuring section there is documentation on how to set the case attribute for rules.  Then each rule that checks case can just reference that section.  I will work on it and commit it.

jeremiah-c-leary:
I just pushed my document updates.  Let me know what you think.

m-kru:
I checked it. I think it is ok.

PR #277, opened 2019-10-16
Add prefix and suffix rules.

m-kru:
I had to break backward compatibility of 2 rules: `signal_008` and `port_011`, but I think it was worth. The proposed solution is more concise and robust, as it allows to enforce both prefixes and suffixes in the same time.


jeremiah-c-leary:
This is awesome.  Thank you.

I agree with breaking the backwards compatibility of signal_008 and port_001.
I will merge the pull request and fix the couple of tests that are failing.

PR #292, opened 2019-11-21
Extract type bug fix.

m-kru:
**Description**
State a clear description of the problem and your solution.

**Screenshots**
If applicable, add screenshots to help explain your solution.

**Additional context**
Add any other context about the problem here including related issue numbers.

### DIFF
--- a/vsg/tests/vsg_utils/test_utils.py
+++ b/vsg/tests/vsg_utils/test_utils.py
@@ -263,6 +263,12 @@ class testExtractFunctions(unittest.TestCase):
         sActual = utils.extract_type_name(oLine)
         self.assertEqual(sExpected, sActual)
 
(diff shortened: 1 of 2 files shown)


jeremiah-c-leary:
Hey m-kru,

I suspect the import change is causing Travis CI to fail.  I will have to look into this before I can merge the pull request.

m-kru:
OK, import line does not work on my machine. I guess you did not want to import PyPI line, but line file from vsg.

jeremiah-c-leary:
Interesting...I will investigate.

jeremiah-c-leary:
I do not understand why there is a difference, but we can resolve it with the following change:

```python
try:
    from vsg import line
except ImportError:
    import line
```

PR #293, opened 2019-11-21
Make instantiation rule_011 inherit from case_rule.

m-kru:
During this change it turned out that there was a bug in how
case was fixed for case rule. It is now fixed, because words that
needs fixing are now stored per line, not per file like it was
previously.


jeremiah-c-leary:
which version of Python are you running?

m-kru:
I use arch, I am always super up to date. Right now Python 3.8.

PR #329, opened 2020-01-22
Fixes required to handle double hyphens in quoted strings so they are not considered a comment marker.

willkamp:
**Description**
The code assumes that anywhere there is a double hyphen (--) that this indicates a comment. This is not entirely true as '-' is also used by std_logic_1164 as the symbol for don't care. Two hyphens within double quotes may indicate a two bit std_logic_vector ("--") or x"0--0" a 16-bit std_logic_vector with 8 bits of don't care.
Double hyphens might also occur in strings (also within double quotes), e.g. "This -- is not a comment."

These code fixes distinguish between double hyphens within a quoted string and comment markers (outside quotes).
The proposed change to tokenisation may make this obsolete.

Couple of other minor fixes that my VHDL code base threw up: dbcebf0, ab97fcb, 3856fc8.

I've also included a few other non-functional changes that make the code more pythonic: 41c3b72, 
3856fc8

### DIFF
--- a/vsg/__main__.py
+++ b/vsg/__main__.py
@@ -103,17 +103,18 @@ def write_invalid_configuration_junit_file(sFileName, commandLineArguments):
 
 
 def write_vhdl_file(oVhdlFile):
-    with open(oVhdlFile.filename, 'w') as oFile:
-        for oLine in oVhdlFile.lines[1:]:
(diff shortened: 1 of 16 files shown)


jeremiah-c-leary:
Greetings Will,

Thank you for the pull request.

PR #338, opened 2020-02-20
Refactor keyword alignment rules [do not merge yet].

m-kru:


jeremiah-c-leary:
I created an example entity declaration with the colons misaligned and had comments and blank lines breaking up groups of ports.  Using the configuration was pretty easy and I could see how the code changes depending on which combination of options you choose.  I will test the other keyword alignment rules, but so far it looks good.  I like how flexible entity_017 is now.  It is no longer just my style.  It now supports other styles without having the user create a new rule on their own.

I noticed the `if_control_statements_end_group` configuration.  Would we need something similar for the `whens` in a case statement?  I will create a testcase.

A tangent thought.  I added the `-oc` option to save the entire configuration VSG is using to a file.  The thought being a user could then use it in future runs, or at least see how the rules were configured.  This is how I found the `if_control_statements_end_group` option.  I was thinking of adding an argument to the option so it would only save/print a single rule.  Then I could just ask VSG what I can configure for any particular rule:
```json
      "entity_017": {
        "blank_line_ends_group": true,
        "comment_line_ends_group": true,
        "disable": false,
        "fixable": true,
        "if_control_statements_end_group": true,
        "indentSize": 2,
        "phase": 5
      },
```
and then use the output to add to my configuration.

m-kru:
> I noticed the if_control_statements_end_group configuration. Would we need something similar for the whens in a case statement? I will create a testcase.

I think we need. I am the stage of fixing failed tests and no such scenario was covered.

> I was thinking of adding an argument to the option so it would only save/print a single rule. 

I really like this idea. By the way I would not add the functionality to save to the file, only print to stdout. Printing is easier, you got less complex application and user can always redirect to file in a terminal.

jeremiah-c-leary:
I came up with this test case:
```vhdl

 architecture RTL of SOME_BLOCK is
 
 begin
 
   PROC_LABEL : process (A, B, C) is
   begin
     case A is
     when A =>
       X <= F;
       XY <= G;
       XYZ <= H;
     when B =>
       x <= I;
       g <= h;
       y <= a;
     when others =>
       null;
   end case;
   end process PROC_LABEL;
 
 end architecture RTL;
```
ends up being fixed to:
```vhdl
 architecture RTL of SOME_BLOCK is
 
 begin
 
   PROC_LABEL : process (A, B, C) is
   begin
 
     case A is
 
       when A =>
         X   <= F;
         XY  <= G;
         XYZ <= H;
       when B =>
         x   <= I;
         g   <= h;
         y   <= a;
       when others =>
         null;
 
     end case;
 
   end process PROC_LABEL;
 
 end architecture RTL;
```

jeremiah-c-leary:
I think this looks really good.  I like the new architecture_026 and process_033.  It makes a lot more sense than the previous signal and constant rules.  Those rules also align the constant colons with the signal colons, which is something people at my work complained VSG did not do.

BTW, I used the **-rc** option to configure those rules to see how I could change them.  It was really helpful.  I had some problem figuring out which rule to use for aligning the colons in the architecture.  Not sure the best way to inform people of which rules applied which changes.

Overall, this is an awesome update.  Thank you very much.

PR #342, opened 2020-02-24
fixed AttributeError

joerg-de:
AttributeError: 'NoneType' object has no attribute 'group' if a file has no signal

**Description**
vsg crashes with "AttributeError: 'NoneType' object has no attribute 'group'" because match returns NoneType if expression is not matched

### DIFF
--- a/vsg/rules/signal/rule_015.py
+++ b/vsg/rules/signal/rule_015.py
@@ -25,7 +25,7 @@ class rule_015(rule.rule):
             self.sFullLine = ''
             self.iFailureLine = iLineNumber
         if oLine.insideSignal:
-            self.sFullLine += oLine.line
+            self.sFullLine += oLine.lineLower
         if oLine.isEndSignal:
             match = re.match(r'.*?signal\s+(?P<signals>[^:\n]*):', self.sFullLine)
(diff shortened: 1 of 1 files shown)


jeremiah-c-leary:
Greetings joerg-de,

Thank you for the pull request.  I wanted to update my tests to catch this error, but I can not seem to replicate the issue.  Is there a chance you could supply an example?

I tried this example with a file without any signals:
```vhdl
 architecture RTL of ENTITY1 is
 
 begin
 
 end architecture RTL;
```
but it does not crash.

joerg-de:
I forgot the main patch. That prevents the crash that happens if a letter in Signal is Upper case

example

` architecture RTL of ENTITY1 is
 
 begin
 
 SIGNAL s : std_logic;
 
 end architecture RTL;
`
the if statement is only to catch invalid code like this ( the : is missing that is what I meant with no signal)

` architecture RTL of ENTITY1 is
 
 begin
 
 SIGNAL s  std_logic;
 
 end architecture RTL;
`

jeremiah-c-leary:
Thanks for the clarification.  I was able to validate the crash.

PR #357, opened 2020-03-21
Issue 344

jeremiah-c-leary:
Updated VSG to handle individual rule configurations per file


jeremiah-c-leary:
Well, that took entirely too long.  Had an issue where Travis CI was returning a dictionary in a different order than my local machine.  I had to write the test so it could handle either case.

Issue #354, opened 2020-03-18
Updates to console output


jeremiah-c-leary:
In a gitter discussion with @nfrancque , it was suggested to add column headers to the output.  This should make it clearer what each column represents.  After reviewing the output, I was wondering if there are other changes that should be made:

- Report configurations
- Point user to documentation on read the docs
- It would be awesome to have the rules "link" to the read the docs website
- Remove the Total Rules Checked and Total Failures?
- Remove the phase reporting?

How can we make the output more user friendly?

jeremiah-c-leary:
I found this and wonder if it would work for links:

echo -e '\e]8;;http://example.com\aThis is a link\e]8;;\a'

If I can get this to work I can link the failure to the readthedocs website

although it would always link to the latest ones

nfrancque:
https://docs.pylint.org/en/1.6.0/output.html describes a little bit of how pylint does their output configuration.  Evidently you can write whatever string with the variables provided and messages come out in that format.

Would it be possible to have a command line option for getting error descriptions?

jeremiah-c-leary:
> Would it be possible to have a command line option for getting error descriptions?

Something like man page?

jeremiah-c-leary:
What do you think of something like this:
```mono
   $ vsg -f PIC.vhd 
   =======================================================================================
   File:  PIC.vhd
   =======================================================================================
   Phase 5 of 7... Reporting
   Total Rules Checked: 208
   Total Violations:    523
   ----------------------------+------------+---------------------------------------------
     Violation                 |  line(s)   | Solution
   ----------------------------+------------+---------------------------------------------
     comment_002               |         51 | Ensure proper alignment of comment with previous line.
     comment_002               |         52 | Ensure proper alignment of comment with previous line.
     comment_002               |         54 | Ensure proper alignment of comment with previous line.
     comment_002               |         55 | Ensure proper alignment of comment with previous line.
     comment_003               |     76-256 | Inconsistent alignment of comments within process.
     sequential_005            |      87-93 | Inconsistent alignment of "<=" in group of lines.
     sequential_005            |    102-103 | Inconsistent alignment of "<=" in group of lines.
     sequential_005            |    105-108 | Inconsistent alignment of "<=" in group of lines.
     sequential_005            |    110-113 | Inconsistent alignment of "<=" in group of lines.
     sequential_005            |    115-118 | Inconsistent alignment of "<=" in group of lines.
[...]

Issue #343, opened 2020-02-25
Setting the default case to lowercase for all rules.


m-kru:
@jeremiah-c-leary I would like to propose something that I know you might not accept, but I want to try as I can see a lot of gains.  Namely, setting the default case to lowercase for all rules.

I am updating the docs for the keyword alignment rules and I see that the structure of the docs could be more consistent and shorter. In case of keyword alignment rules all default configs are the same, so it is documented only once. In case of case rules you can see following line for each of them:
```
.. NOTE:: The default is lowercase/uppercase.
```
If all case rules were lowercase it would be much easier for the user to know the default configurations (more user friendly). What is more the documentation would be shorter as well as the python codes (less characters to maintain). Violations and fixes examples would be more consistent as everything would be lowercase. 

VHDL is case insensitive anyway, and I know people who do not use uppercase at all because of that. It is easier to type without Shift/Caps Lock.

Some backward compatibility is already broken because some of the rules have been removed and some others changed their behavior a little, so it would be good time to introduce such change.
What is more, right now we have to care about the backward compatibility because some of the configurations are hard coded as a defaults. I think it would be nice to provide users with some predefined configurations. They could be placed for example in the `example_configurations`  directory in the project. Thanks to this, it would be easier to provide backward compatibility as in case of any changes in the code the example configurations could be adjusted.

jeremiah-c-leary:
@m-kru I am not apposed to setting all the cases to lower.  Those rules which default to `upper` are just arbitrary because that is how I prefer them to be set.

>I think it would be nice to provide users with some predefined configurations. They could be placed for example in the example_configurations directory in the project. Thanks to this, it would be easier to provide backward compatibility as in case of any changes in the code the example configurations could be adjusted.

I agree with this.  I was struggling with how to keep "backwards compatibility" and I think this would solve that issue.  It would be really awesome if people would add their own configurations.

Maybe add another command line option for built in configurations?

```bash
   $ vsg -bic legacy
   $ vsg -bic compact
   $ vsg -bic testbench
   $ vsg -bic rtl
   $ vsg -bic <company_name>
```
the option could search a directory for configuration.

We could also add a `description` option into the configuration:

```json
   {
     "name" : "legacy"
     "description" : [
         "This is the configuration which matches the original setup for VSG."
      ],
     "rule": {
     }
  }
```

Then if you issue the `bic` option without anything you would get a report out:

```bash
   $ vsg -bic
    Available build in configurations:
[...]

PR #358, opened 2020-03-25
Issue 354

jeremiah-c-leary:
This pull request is the implementation of issue #354.  The output has change to:

``` bash
   $ vsg -f PIC.vhd 
   =======================================================================================
   File:  PIC.vhd
   =======================================================================================
   Phase 5 of 7... Reporting
   Total Rules Checked: 208
   Total Violations:    523
   ----------------------------+------------+---------------------------------------------
     Violation                 |  line(s)   | Solution
   ----------------------------+------------+---------------------------------------------
     comment_002               |         51 | Ensure proper alignment of comment with previous line.
     comment_002               |         52 | Ensure proper alignment of comment with previous line.
     comment_002               |         54 | Ensure proper alignment of comment with previous line.
     comment_002               |         55 | Ensure proper alignment of comment with previous line.
     comment_003               |     76-256 | Inconsistent alignment of comments within process.
     sequential_005            |      87-93 | Inconsistent alignment of "<=" in group of lines.
     sequential_005            |    102-103 | Inconsistent alignment of "<=" in group of lines.
     sequential_005            |    105-108 | Inconsistent alignment of "<=" in group of lines.
     sequential_005            |    110-113 | Inconsistent alignment of "<=" in group of lines.
     sequential_005            |    115-118 | Inconsistent alignment of "<=" in group of lines.
     sequential_005            |    120-124 | Inconsistent alignment of "<=" in group of lines.
     sequential_005            |    129-133 | Inconsistent alignment of "<=" in group of lines.
     sequential_005            |    160-161 | Inconsistent alignment of "<=" in group of lines.
     sequential_005            |    173-174 | Inconsistent alignment of "<=" in group of lines.
     comment_002               |        183 | Ensure proper alignment of comment with previous line.
     sequential_005            |    225-226 | Inconsistent alignment of "<=" in group of lines.
     sequential_005            |    238-239 | Inconsistent alignment of "<=" in group of lines.
   ----------------------------+------------+---------------------------------------------
```

### DIFF
--- a/docs/example.rst
+++ b/docs/example.rst
@@ -1,0 +1,76 @@
+
+.. code-block:: bash
+
+   $ vsg -f PIC.vhd 
+   =======================================================================================
+   File:  PIC.vhd
+   =======================================================================================
+   Phase 5 of 7... Reporting
+   Total Rules Checked: 208
+   Total Violations:    523
+   ----------------------------+------------+---------------------------------------------
+     Violation                 |  line(s)   | Solution
+   ----------------------------+------------+---------------------------------------------
+     comment_002               |         51 | Align comment with previous line. 
+     comment_002               |         52 | Align comment with previous line. 
+     comment_002               |         54 | Align comment with previous line. 
+     comment_002               |         55 | Align comment with previous line. 
+     comment_003               |     76-256 | Align comments within process.
+     sequential_005            |      87-93 | Align "<=" within the group of lines.
+     sequential_005            |    102-103 | Align "<=" within the group of lines. 
+     sequential_005            |    105-108 | Align "<=" within the group of lines. 
+     sequential_005            |    110-113 | Align "<=" within the group of lines. 
+     sequential_005            |    115-118 | Align "<=" within the group of lines. 
+     sequential_005            |    120-124 | Align "<=" within the group of lines. 
+     sequential_005            |    129-133 | Align "<=" within the group of lines. 
+     sequential_005            |    160-161 | Align "<=" within the group of lines. 
+     sequential_005            |    173-174 | Align "<=" within the group of lines. 
+     comment_002               |        183 | Align comment with previous line.
+     sequential_005            |    225-226 | Align "<=" within the group of lines. 
+     sequential_005            |    238-239 | Align "<=" within the group of lines. 
+   ----------------------------+------------+---------------------------------------------
+   NOTE: Refer to online documentation at http://... for more information.
+
+
+   $ vsg -f PIC.vhd 
+   =======================================================================================
+   File:  PIC.vhd
+   =======================================================================================
+   Phase 7 of 7... Reporting
+   Total Rules Checked: 208
+   Total Violations:      0
+
+   
+
+
+   $ vsg -f PIC.vhd 
+   File:  PIC.vhd
+   ==============
+   Phase 1... Reporting
+   Phase 2... Reporting
+   Phase 3... Reporting
+   Phase 4... Reporting
+   Phase 5... Reporting
+     comment_002               |         51 | Ensure proper alignment of comment with previous line.
+     comment_002               |         52 | Ensure proper alignment of comment with previous line.
+     comment_002               |         54 | Ensure proper alignment of comment with previous line.
+     comment_002               |         55 | Ensure proper alignment of comment with previous line.
+     comment_003               |     76-256 | Inconsistent alignment of comments within process.
+     sequential_005            |      87-93 | Inconsistent alignment of "<=" in group of lines.
+     sequential_005            |    102-103 | Inconsistent alignment of "<=" in group of lines.
+     sequential_005            |    105-108 | Inconsistent alignment of "<=" in group of lines.
+     sequential_005            |    110-113 | Inconsistent alignment of "<=" in group of lines.
+     sequential_005            |    115-118 | Inconsistent alignment of "<=" in group of lines.
+     sequential_005            |    120-124 | Inconsistent alignment of "<=" in group of lines.
+     sequential_005            |    129-133 | Inconsistent alignment of "<=" in group of lines.
+     sequential_005            |    160-161 | Inconsistent alignment of "<=" in group of lines.
+     sequential_005            |    173-174 | Inconsistent alignment of "<=" in group of lines.
+     comment_002               |        183 | Ensure proper alignment of comment with previous line.
+     sequential_005            |    225-226 | Inconsistent alignment of "<=" in group of lines.
+     sequential_005            |    238-239 | Inconsistent alignment of "<=" in group of lines.
+   Phase 6... Not executed
+   Phase 7... Not executed
+   ==============
+   Total Rules Checked: 204
+   Total Failures:      523
+
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -101,6 +101,7 @@ Here is an example output running against a test file:
      signal_005                |          4 | Ensure only a signal space after the colon.
      signal_005                |          5 | Ensure only a signal space after the colon.
    ----------------------------+------------+--------------------------------------
+   NOTE: Refer to online documentation at https://vhdl-style-guide.readthedocs.io/en/latest/index.html for more information.
 
 VSG will report the rule which is violated and the line number or group of lines where the violation occured.
 It also gives a suggestion on how to fix the violation.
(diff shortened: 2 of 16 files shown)


docs/usage.rst
```diff
@@ -87,36 +87,21 @@ Here is an example output running against a test file:
...
-   Total Failures:      523
+   $ vsg -f example/architecture-empty.vhd
+   ================================================================================
+   File:  example/architecture-empty.vhd
+   ================================================================================
+   Phase 2 of 7... Reporting
+   Total Rules Checked: 151
+   Total Violations:    3
+   ----------------------------+------------+--------------------------------------
+     Violation                 |  line(s)   | Solution
```
m-kru: Shouldn't the name of the first column be Rule not Violation?

docs/example.rst
```diff
@@ -0,0 +1,76 @@
...
+
+   $ vsg -f PIC.vhd 
+   =======================================================================================
+   File:  PIC.vhd
+   =======================================================================================
+   Phase 5 of 7... Reporting
+   Total Rules Checked: 208
+   Total Violations:    523
+   ----------------------------+------------+---------------------------------------------
+     Violation                 |  line(s)   | Solution
```
m-kru: Shouldn't the name of the first column be Rule not Violation?

docs/usage.rst
m-kru: The title of the whole table should probably be Violations.

docs/example.rst
m-kru: The title of the whole table should probably be Violations.

jeremiah-c-leary: You are correct.  Violation should be Rule.

jeremiah-c-leary:
> I got a feeling that `utils.create_violation_dict()` is unnecessary. The violation list stores violations for given list. If for particular rule the the only thing that needs to be stored is the `lineNumber`, then it is ok to store the list of integers.

I had considered keeping the integers, but I thought having a consistent entry point into a dictionary would be better.  For the simple cases it is overkill though.

m-kru:
`utils.create_violation_dict()` makes assumption that argument passed to it is a `lineNumber`. What in case of other rules that have single key in violation? Do we assume that if rule has only one key in the violation, then it must be `lineNumber`?
  

> I thought having a consistent entry point into a dictionary would be better.

Is there any use case for having consistent entry point?

jeremiah-c-leary:
> Is there any use case for having consistent entry point?

This is a little rambly, but I eventually answer your question...

When we switched from putting integers into `vsg.rule.violations` to allowing dictionaries, we allowed each rule to decide how to store violations.  This made sense at the time, but I did not realize there was an issue when reporting the violations.

In `vsg.rule_list.report_violations` method it eventually calls the `vsg.rule.report_violations` method.  
It was responsible for checking if a violation occurred at a specific line number, because I wanted to print the violations in sequential order, and generating the output.

When we allowed dictionaries, things became a little more complex.  My updates to dictionaries resulted in this solution to the `report_violations` method:
```python
         for violation in self.violations:
             if isinstance(violation, dict):
                 sViolation = str(violation['lineNumber'])
             else:
                 sViolation = violation
```
I used the else to maintain backwards compatibility with users who created their own rules.

After I merged your final keyword alignment changes, I was testing some example files and everything was working.
```vhdl

 entity FIFO is
   port (
     I_PORT1  : in    std_logic;
     I_PORT2 : in    std_logic;
     O_PORT1 : out   std_logic
   );
 end entity FIFO;
```
VSG was fixing the `:` alignment in my entity when using the `--fix` option.  So I decided to check the reporting and it crashed.
```mono
vhdl-style-guide$ python3 bin/vsg -f examples/entity.vhd
File:  examples/entity.vhd
==========================
Phase 1... Reporting
Phase 2... Reporting
Phase 3... Reporting
Phase 4... Reporting
Phase 5... Reporting
Traceback (most recent call last):
  File "bin/vsg", line 14, in <module>
    main()
  File "/home/jeremiah/projects/vhdl-style-guide/bin/../vsg/__main__.py", line 291, in main
    oRules.report_violations(commandLineArguments.output_format)
  File "/home/jeremiah/projects/vhdl-style-guide/bin/../vsg/rule_list.py", line 179, in report_violations
    iFailures += oRule.report_violations(iLineNumber, sOutputFormat, self.oVhdlFile.filename)
  File "/home/jeremiah/projects/vhdl-style-guide/bin/../vsg/rule.py", line 44, in report_violations
    sViolation = str(violation['lineNumber'])
KeyError: 'lineNumber'

```
While debugging the crash I realized the `keyword_alignment` rule had a violation dictionary structure:
```json
   "lines":[],
   "max_keyword_column":10,
   "max_before_keyword_column":6
```
Versus what I had implemented:
```json
   "linenumber":10
   "other_key":"some_string_or_number"
```
Thinking about the differences, it made sense.  Most of the rules deal with a single line.  The keyword alignment rule needs to gather multiple lines and then compare the keyword in each line.  The violation is on the group of lines, but you need to store the information for each line within the group.

So in deciding how to handle this, I wondered why it is the rule classes responsibility to format data for stdout.  Which is why there is now a `vsg.report` module.  Then I replaced `vsg.rule.report_violation` method with `vsg.rule.get_violation_at_linenumber`, which looks like:

```python
     def get_violations_at_linenumber(self, iLineNumber):
         '''
         Returns a list of formatted violations.
 
         Parameters:
 
           iLineNumber (integer)
 
         Returns: (list of dictionaries)
         '''
         lReturn = []
 
         for violation in self.violations:
             check_for_old_violation_format(violation)
             lKeys = list(violation.keys())
             if 'lineNumber' in lKeys:
                 sViolation = str(violation['lineNumber'])
                 self._build_violation_dict(lReturn, sViolation, iLineNumber)
 
             elif 'lines' in lKeys:
                 for dLineViolation in violation['lines']:
                     sViolation = str(dLineViolation['number'])
                     self._build_violation_dict(lReturn, sViolation, iLineNumber)
         return lReturn
```
Now VSG can handle the `keyword_alignment` format and the other format.  I switched all the other violations to use a dictionary instead of a number.  I also added  check for integers as a violation and print out an ERROR and point users to some documentation on using a dictionary.

So...this is long winded explaination that doesn't quite answer your question, but I wanted to show you how I ended up with my solution.

To answer your question explicitly:  I think consistency only matters when we are reporting out violations.  Interestingly, each rule is independent when fixing violations.

m-kru:
@jeremiah-c-leary I understand.
Maybe we could put a constraint on dictionaries that are passed to the `add_violation()` function? Force that each such instance of dictionary must have `lines` attribute. In case of single line rules `lines` would be an one element array. Then instead of:
```python
             if 'lineNumber' in lKeys:
                 sViolation = str(violation['lineNumber'])
                 self._build_violation_dict(lReturn, sViolation, iLineNumber)
 
             elif 'lines' in lKeys:
                 for dLineViolation in violation['lines']:
                     sViolation = str(dLineViolation['number'])
                     self._build_violation_dict(lReturn, sViolation, iLineNumber)
```
we would have:
```python
             for dLineViolation in violation['lines']:
                 sViolation = str(dLineViolation['number'])
                 self._build_violation_dict(lReturn, sViolation, iLineNumber)
```

jeremiah-c-leary:
> It does not make much sense to maintain both lineNumber and lines, when lineNumber is just a subset of lines case.

I agree.  I was struggling with how to combine the two, and I like your approach.  i will check to see what it will take to convert to that format.

m-kru:
I have come across the same issue:
```
  File "/home/mkru/workspace/vhdl-style-guide/vsg/rule.py", line 44, in report_violations
    sViolation = str(violation['lineNumber'])
KeyError: 'lineNumber
```
while working on https://github.com/jeremiah-c-leary/vhdl-style-guide/issues/343 .

When do you plan to finish this pull request?

jeremiah-c-leary:
> When do you plan to finish this pull request?

If things go well, it should be today.

m-kru:
@jeremiah-c-leary I think we may need one more issue, namely to make rules using `lineNumber` start using `lines[]`.

jeremiah-c-leary:
> I think we may need one more issue, namely to make rules using lineNumber start using lines[].

I think that makes sense.  Issue 354 uncovered the discrepancy, but that doesn't mean it should be the vehicle to fix it.  The reporting is done.

I will create another issue and work on that one.  When that one is complete, then I can merge in this pull request.

m-kru:
> I will create another issue and work on that one. When that one is complete, then I can merge in this pull request.

Great!

jeremiah-c-leary:
@m-kru I was just thinking...I could merge this pull request in, which will allow you to report.  Then I can use issue #362 to simplify some of the code introduced during this commit.

Let me know if you agree.

m-kru:
@jeremiah-c-leary it would be good if I could continue working as I have some spare time today.

Issue #362, opened 2020-03-29
Unify violation dictionaries


jeremiah-c-leary:
Currently, there are two styles of violation dictionaries:
```json
  {
    "lineNumber":20
  }
```
...and
```json
  {
    "lines":[
     {
       "number":20
     }
    ]
  }
```
This issue was discovered while implementing issue #354  
The consensus is to use the later format.

The definition of done includes:

[ ] document violation dictionary format
[ ] update rules which use the former style to use the later style
[ ] all tests passing

PR #363, opened 2020-03-29
Updates for issue #362:

jeremiah-c-leary:
I wanted to document to violation dictionary to ensure we have the same understanding of it before I implement any changes.

### DIFF
--- a/docs/localizing.rst
+++ b/docs/localizing.rst
@@ -403,6 +403,95 @@ Finally, we update the line with our modified line using the **update_line** met
 
                 oFile.lines[iLineNumber].update_line(sNewLine)
 
+Violation dictionary
+--------------------
+
+Violations are stored as a list of dictionaries in the **rule.violations** attribute.
+This is the generic format of the dictionary represented by json:
+
+.. code-block:: json
+
+   {
+     "lines" : [ 
+        {
+          "number" : "<integer>",
+          "<line_attribute>" : "<line_value>",
+          "<line_attribute>" : "<line_value>"
+        }
+       ],
+     "<violation_attribute>" : "<violation_value>",
+     "<violation_attribute>" : "<violation_value>"
+   }
+
+This format gives us the greatest flexibility in describing violations.
+The lines[0]['number'] is the only required element in a violation dictionary.
+The "<line_attribute>" and "<violation_attribute>" elements are optional.
+They are used by more complex rules to maintain information used to fix violations.
+
+Single line violations
+~~~~~~~~~~~~~~~~~~~~~~
+
+Most violations are against a single line and no other information is required to fix it.
+These dictionaries use the minimumal form.
+
+.. code-block:: json
+
+   {
+     "lines" : [ 
+        {
+          "number" : 40
+        }
+       ]
+   }
+
+Single line violations with additional information
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If additional information for single line violations is required, it will be stored at the **violation** level.
+
+.. code-block:: json
+
+   {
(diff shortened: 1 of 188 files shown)


docs/localizing.rst
```diff
@@ -403,6 +403,95 @@ Finally, we update the line with our modified line using the **update_line** met
...
+     "lines" : [ 
+        {
+          "number" : "<integer>",
+          "<line_attribute>" : "<line_value>",
+          "<line_attribute>" : "<line_value>"
+        }
+       ],
+     "<violation_attribute>" : "<violation_value>",
+     "<violation_attribute>" : "<violation_value>"
+   }
```
m-kru: Maybe we should keep `lines` as a pure integer array with line numbers. not an array of dictionaries? If we need more data associated with given line we could always add another list or lists and associate data by index (lists are ordered).
This would make `lines` attribute more clear for rules that need only line numbers. No need to have `number` key.

jeremiah-c-leary: what I like about the format is it keeps the information atomic.  So information about the line is kept together and information about all the lines are kept together.

The keyword alignment rule would create a violation dictionary like this:

```json
{
  "lines" : [
    {
      "number" : 31,
      "keyword_column" : 52,
      "before_keyword_column" : 12
    },
    {
      "number" : 32,
      "keyword_column" : 45,
      "before_keyword_column" : 6
    },
    {
      "number" : 32,
      "keyword_column" : 76,
      "before_keyword_column" : 30
    }
  ],
  "max_keyword_column" : 76,
  "max_keyword_before_column" : 30
}
```
We can always write a function to pull out all the line numbers:

```python
def extract_line_numbers_from_violation_dict(dViolation):
    lReturn = []
    for dLine in dViolation['lines']:
        lReturn.append(dLine['number'])
    return lReturn
```
it seems like overkill for single line violations, but I have started using these functions:

```python
def create_violation_dict(iLineNumber):
    '''
    Builds a minimal violation dictionary.

    Parameters:

      iLineNumber: (integer)

    Returns:  dictionary
    '''
    dReturn = {}
    dReturn['lines'] = []
    dLine = {}
    dLine['number'] = iLineNumber
    dReturn['lines'].append(dLine)
    return dReturn


def get_violation_linenumber(dViolation):
    '''
    Returns a line number of a violation.

    Parameters:

      dViolation: Violation dictionary

    Returns:  integer
    '''
    return dViolation['lines'][0]['number']


def  get_violating_line(oFile, dViolation):
     '''
     Returns a line from the file where a violation has occured.

     Parameters:

       oFile : (File object)

       dViolation : (Violation dictionary)

     Return: Line Object
     '''
     return oFile.lines[get_violation_linenumber(dViolation)]
```
To help abstract away the structure of the dictionary.

m-kru: Well, no matter which way we choose, it would be more coherent than what we have right now.

jeremiah-c-leary:
@m-kru I am going to merge this in tomorrow.  I wanted to give you a heads up in case there was anything I should consider before I merge it.

vsg/rules/concurrent/rule_007.py
```diff
@@ -32,7 +32,7 @@ def _analyze(self, oFile, oLine, iLineNumber):
 
     def _fix_violations(self, oFile):
         for dViolation in self.violations[::-1]:
-            iLineNumber = dViolation['lineNumber']
+            iLineNumber = utils.get_violation_linenumber(dViolation)
```
m-kru, line 35: Shouldn't it be `line_number`, not `linenumber`?

jeremiah-c-leary: I always struggle with how to name things.  I think that makes sense.  I will make the update.

PR #366, opened 2020-03-29
Set the default case to lowercase for all case rules.

m-kru:

### DIFF
--- a/docs/configuring_case.rst
+++ b/docs/configuring_case.rst
@@ -2,17 +2,17 @@ Configuring Uppercase and Lowercase Rules
 -----------------------------------------
 
 There are several rules that enforce either uppercase or lowercase.
-They are noted in the documentation for each rule what the default is.
-The example violation and fixes are in reference to the default setting.
-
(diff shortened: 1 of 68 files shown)


docs/configuring_case.rst
```diff
@@ -22,22 +22,11 @@ We can use the following configuration to change the case to upper:
...
-   rule :
-     entity_008 :
-        case : 'lower'
-
 Changing Multiple Case Rules
 ############################
 
 If there are a lot of case rules you want to change, you can use the global option to reduce the size of the configuration.
-For example, if we want to uppercase everything except the entity name, we could write the following configuration:
+For example, if you want to uppercase everything except the entity name, you could write the following configuration:
```
jeremiah-c-leary, line 29: It is interesting I use **we** instead of **you**

m-kru: Personally, I have never come across 'we' in any documentation or tutorial. In humble opinion it is better to use 'you' when the software is provided as is, without warranty of any kind.

Issue #355, opened 2020-03-18
Allow user to augment existing command line arguments


jeremiah-c-leary:
In a gitter discussion with @nfrancque it was proposed to add the ability for end users to augment command line arguments to support scripting.

This would involve being able to request the command line arguments from VSG similar to how VUNIT works.

This issue will discuss how this should work, as it is still a little unclear to me, and whether to implement it.

nfrancque:
So at the time of suggesting this I was unaware the config file could also contain command line arguments. That is an interesting approach I have not seen before, and means this is definitely less valuable than with vunit. 
If that is the intended approach I am not sure most users are aware of that ability, is it featured in the examples? 

The only thing this feature would then add is if a user wanted to also do some extra work in front of vsg. You can imagine any number of use cases - gathering up what files to give vsg, pre or post processing vsg results, adding optional simplified vsg arguments - in general building domain specific tool variants on top of vsg.

jeremiah-c-leary:
Hey @nfrancque ,

I am going through old issues and was wondering if this is still something you would like to have.

--Jeremy

jeremiah-c-leary:
Hey @nfrancque ,

I am going to set to defer and schedule to close it after Dec 31st.

--Jeremy

PR #367, opened 2020-04-02
Updates for issue #353:

### LINKED ISSUE #353: Default configurations
A discussion on gitter with @nfrancque proposes having a layered configuration approach where the user can run with default configurations without having to use the `--configuration` option.  The proposal includes:

[ ] Having a user defined configuration stored in their home account stored as a `.vsgrc` file
[ ] Having a project based configuration possibly set using an environment variable `VSG_PROJECT_CONFIG`

The current proposal would be:

1. use the `.vsgrc` file if it exist and the environment variable `VSG_PROJECT_CONFIG` is not set.
2. Use the configuration defined by the environment variable `VSG_PROJECT_CONFIG` if it is set and ignore the `.vsgrc` file if it exists

We need to determine the interaction of the `--configuration` option when either the `.vsgrc` file or the `VSG_PROJECT_CONFIG` file exists.

I propose `--configuration` has highest priority and we ignore the `.vsgrc` and environment variable.

VSG should also print out which configurations it used.  This should make it clearer to the user when trying to debug why VSG made changes.

jeremiah-c-leary:
Before implementing issue #353 I would like to know if this is the direction we should go and if it solves any issues for the user.

Resolves #353


nfrancque:
So here's how I'm thinking of this:
| Use Case                     | Covered?                     |
| -----------------------  | ------------------------ |
| Project Level Config    | :heavy_check_mark:     |
| Default User Config     | :heavy_check_mark:     |
| Mixed Style Codebase | :x:                                |
| Multiple Projects         | :x:                                 |

We are definitely good in this approach on the first two, but without auto-discovery at the directory level supporting a project with mixed styles will be very cumbersome, and users working on multiple projects at the same time are very likely to end up with a style guide stuck as their default when they move to a project that doesn't have VSG set up.

That isn't inherently a bad thing, but I'd say it's not expected behavior.  

And obviously having a mixed style codebase is undesirable, but realistically it will probably always be the rule rather than the exception.

I want to be able to come in to a project and already be able to run VSG on the IP I've brought in without having to set anything up - ideally, just `vsg .` on the submodules/externals directory and have it find all the configurations of each IP block.  This gives you a good idea of where all the IP is at relative to its own rules, which importantly will probably not be your own.

From a usability perspective, IMO auto-discovery of the most low level config is ideal, but no idea what the implementation complexity might be.

m-kru:
@nfrancque @jeremiah-c-leary In my humble opinion we get no gain with this feature, but add extra complexity. 

For example, here is how I use `vsg`. In each project I have a bash script wrapper for `vsg`. In the bash script I have list of `.vhd` files that should be checked by the `vsg` and a loop for checking them. With such approach it is easy to have multiple lists and even check them with different configurations, for example different list for testbench files. Adding new file for checking is super easy, you just add the file to the proper list in bash script. There are 2 more advantages:

- it is super easy to integrate it with CI/CD server. You just run the script on the server.

- sometimes I import files to the project with absolutely different formatting rules. If I do not want to check them against `vsg`, I just do not add them to the file lists.

@nfrancque Would such approach suit you?

nfrancque:
I think that can work, but you are exponentially increasing the odds of crappy IP existing by not including it in your project VSG runs. By making sure in every project the IP adheres to its own rules bugs will be found much more quickly. 

Manually searching for a script is also prone to missing something. For larger projects they will probably forget to add a block or two, and unlike compiles there will be nothing telling you you have forgotten to add something. Have seen that at work already, actually.

m-kru:
> I think that can work, but you are exponentially increasing the odds of crappy IP existing by not including it in your project VSG runs. By making sure in every project the IP adheres to its own rules bugs will be found much more quickly.

No, I do not increase anything. IP comes from some library, each library may have different formatting rules. I do not want to adjust each of the imported files to a given project rules.
If imported IP is formatted in a really crappy way (no formatting at all), then it is the problem of the library, and it should be fixed there, not in my project.

> Manually searching for a script is also prone to missing something. For larger projects they will probably forget to add a block or two, and unlike compiles there will be nothing telling you you have forgotten to add something. Have seen that at work already, actually.

Then inverse the logic. In bash script find all `.vhd` files and make your list to contain files that should NOT be checked. You will never have a situation like _forget to add something_.

nfrancque:
This is interesting, I think we are saying similar things in different ways 😃 I agree adjusting the rules is a no go, but want to account for that by having vsg automatically switch configurations based on the most local rule set. This way if the ip does not follow its own rules it is indeed your problem. You may choose to ignore that later but it should be a conscious decision. Does that make sense? Otherwise it will simply never be fixed. 

On the latter point, I am referring to the aggregation of block level vsg scripts. Inevitably someone will forget to add a block. 

I think my thoughts here are kind of outside the scope of this issue / pull request but definitely related and would, as you have said, make this pull request of lesser value.

m-kru:
> On the latter point, I am referring to the aggregation of block level vsg scripts.

Like different scripts in the subdirectories?

I just feel like we are trying to fix something that is not a `vsg` problem per se and in my humble opinion it should be handled outside the `vsg`.

nfrancque:
Understood.  I don't agree with that personally but could absolutely see it going both ways.

IMO, the tool's goal should be kept in mind - if it is "Find and optionally fix all style issues in a **project**", then things like this are well within scope.  If the goal is more "Find and optionally fix all style issues in a **file** with some support for multiple files", then that is different, and your opinion is most likely the correct one.

This is also related to to #355 in that it would be very necessary to add its functionality if VSG itself does not address project level options like this.  And realistically, I would probably add the above functionality to a tool wrapping VSG if #355 is implemented and this is not 😄

m-kru:
> IMO, the tool's goal should be kept in mind - if it is "Find and optionally fix all style issues in a project", then things like this are well within scope. If the goal is more "Find and optionally fix all style issues in a file with some support for multiple files", then that is different, and your opinion is most likely the correct one.

This is interesting, as I see the main goal of the vsg project to: "Find and optionally fix all style issues in a **file**". This may be because I really like the Unix philosophy: _Do One Thing and Do It Well_.

Of course the goal of the project should be probably defined by @jeremiah-c-leary as it is his project   :).

nfrancque:
Agree, Unix philosophy is good but someone has to do the project level stuff eventually.  If you expose the internal API then VSG can separately be both the file fixer and the project fixer while adhering the the Unix philosophy.

And yeah, waiting on your opinion, @jeremiah-c-leary !

jeremiah-c-leary:
@m-kru, @nfrancque :  This is great discussion.

> From a usability perspective, IMO auto-discovery of the most low level config is ideal, but no idea what the implementation complexity might be.

Auto-discovery is an interesting concept.  It reminds me of python's unittest where you must follow a defined pattern for it to work.  I could see defining `vsg.yaml` files throughout the directory structure.  VSG could then search for them and execute based on their contents.

I started thinking through any possible issues with this approach and landed on this one:  How would VSG handle configurations for IP which are from a previous version of VSG.

At some point, IP should be stable and it's formatting is locked down.  We are making some rule changes for version 2.0.0.  If we ran a 1.2.0 configured file with 2.0.0 rules  there could be new violations identified.  To handle this, VSG would have to:

1.  know how to handle files from every version
2.  would have to told which version of VSG the IP was run against

The first would require significant effort.  We would have to keep track of which rules exist for each version.  Keep track of changing behaviors, which rules were renamed, which rules were deleted, etc...

The second would require the user to set some field in the configuration to the correct version number.  Granted, that could be done by VSG but it would still require someone to commit any changes to the configuration file.

> I just feel like we are trying to fix something that is not a vsg problem per se, and in my humble opinion it should be handled outside the vsg.

> IMO, the tool's goal should be kept in mind - if it is "Find and optionally fix all style issues in a project", then things like this are well within scope. If the goal is more "Find and optionally fix all style issues in a file with some support for multiple files", then that is different, and your opinion is most likely the correct one.

VSG was always intended to be file centered.  Any project level decisions:  how the files should be formatted or which files to format were left for the user.  I believe this allows the user the greatest flexibility in defining how to use VSG with their project.

For example, @m-kru used a bash script.  I have used configurations on a per block basis in Atlassian's Bamboo CI server.  I'm sure someone else came up with a different solution.

VSG is just one tool out of many that can be used in VHDL development.  What is required is an orchestrator to pull everything together.  Ohhh...that would be a good name for a tool:  Conductor.  If you think of all the tools required to build an FPGA, they are the individual instruments.  While each musician is an expert in their instrument, it is the conductor that makes them work as a team to produce a symphony.  Is there an open source tool that will do this?  I have written two for my company that I do not believe I can open source.

Anyway...back to this discussion.

Conclusion:

Autodetection should be handled at a level higher than VSG.

Question:

If we can alias and wrap VSG with bash, is this feature even needed?

m-kru:
> At some point, IP should be stable and it's formatting is locked down. We are making some rule changes for version 2.0.0. If we ran a 1.2.0 configured file with 2.0.0 rules there could be new violations identified. To handle this, VSG would have to:
> 
> 1. know how to handle files from every version
> 2. would have to told which version of VSG the IP was run against

Any of the two options is plenty of wearisome work. Just to gain something that can be easily solved  'above' the vsg. Besides this, we are going to have example configurations from release 2.0.0 to keep backward compatibility in the future.

> VSG is just one tool out of many that can be used in VHDL development. What is required is an orchestrator to pull everything together. Ohhh...that would be a good name for a tool: Conductor. If you think of all the tools required to build an FPGA, they are the individual instruments. While each musician is an expert in their instrument, it is the conductor that makes them work as a team to produce a symphony. Is there an open source tool that will do this? I have written two for my company that I do not believe I can open source.

I use [FuseSoc](https://github.com/olofk/fusesoc). I was also thinking about adding `vsg` as one of supported back ends, it already supports some lint only tools. I have also written my own tool [fsva](https://github.com/m-kru/fsva) to automatically extract and run test bench targets from FuseSoc. If FuseSoc supported `vsg` I would also automatically extract and run lint targets, before test bench targets. What I like the most in my way of managing the project is the fact, that what I do locally on my machine is exactly the same what I set up on the CI/CD server (I mean literally the same commands).

jeremiah-c-leary:
Finally getting back to this.  After re-reviewing everything, I think the following makes sense:

  1)  Search for configuration file in directory ~/.config/vhdl-style-guide
  2)  If that does not exist then check if environment variable $VSG_DEFAULT_CONFIG is defined

If both 1 and 2 are true, then 1 will be ignored.

I have updated the documentation to reflect this.

m-kru:
@jeremiah-c-leary can you remind what is the idea for it right now. I guess you are no going to implement auto-discovery, are you?

jeremiah-c-leary:
@m-kru I will not be implementing auto-discovery.  VSG will just check if a config file exists in your home directory and the environment variable.

If you have neither, then VSG operates as it did before.  Essentially, they are not required.

m-kru:
@jeremiah-c-leary ok. I still think we are over-engineering. I do not see how it is different from setting an alias in `bash` (or whatever shell the user uses) and putting the .config file wherever you want, even on server and `wget`'ing it.

jeremiah-c-leary:
@m-kru 
> I still think we are over-engineering.
I do not disagree with that statement.  I do not like putting in features nobody is going use.
I will think again about this issue.

jeremiah-c-leary:
okay...so I basically implemented this and ran into an issue with multiply defined file lists.
In the program configuration file I allowed `filename` to be defined, because....it seemed reasonable at the time.  However, there is a `file_list` option in the rule configuration file.  Now something interesting happens when I use a rule configuration.  I setup the code to ignore the `~/.config/vhdl-style-guide/config.yaml` and the one defined by the environment variable if a configuration was passed by the `-c` option.  Now all the options in the program configuration are ignored.  We seem to be back to square one.

So then I decided to review the options which should be allowed to be configured via the program configuration file:

-h : obviously not
-f : should not be
-lr : could be, but could conflict with rule configuration
-c : could be
--fix : I wouldn't want it to be
-fp : should not be
-j : seems overkill to always generate one
-b : I suppose, but it would leave a bunch of files around
-oc : this is a rarely used option
-rc : this is only used when you want to figure out how a rule can be configured
-v : this is only used when you want to know the version
-of : I could see this one being configured

So we are left with three, at the moment, that would make sense to have configurations for.
The `-lr` option would be ignored if a rule configuration is used.
The `-c` option you could have  default config that would be ignored if you passed a rule configuration to it.
The `-of` option makes sense, but if so then you can just use an alias.

My conclusion at the moment is to defer this item and work on others for the 2.0.0 release.

m-kru:
Well, you know my opinion on this issue, useless, over engineered, complicated.

jeremiah-c-leary:
Evening,

Well...it has been about 4 years and I am going to retire this PR and the associated issue.  In that time I have come to the conclusion that VSG should not be responsible for managing configurations.  It will accept them, but the decision on which to use should be determined by the user.

There has not been any traction on this PR or the link issue,.  Therefore I will assume others agree.
 
Regards,

--Jeremy

PR #387, opened 2020-05-08
fixed hang if many quotes are used in a single line

joerg-de:
**Description**
If to many quotes are used in one line vsg hangs this fixes that by replacing the regex with python code

**Test**

""|""|""|""|""|""|""|""|""|""|""|""|""|""|""|""|""|""|""|""|""|""|""|""|

### DIFF
--- a/vsg/utils.py
+++ b/vsg/utils.py
@@ -298,10 +298,8 @@ def search_for_and_remove_keyword(oFile, iLineNumber, sKeyword):
         if not oLine.isBlank:
             break
 
-# Regex to find comments that ignores contents of double quoted strings,
+# Find comments and ignores contents of double quoted strings,
 # for example, "--" : a two bit std_logic_vector literal of don't cares.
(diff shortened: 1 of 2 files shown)


jeremiah-c-leary:
Hey @joerg-de, could you provide a little more context for the error.  A snippet of an example code that causes the issue would be very helpful

joerg-de:
the code that crash is ""|""|""|""|""|""|""|""|""|""|""|""|""|""|""|""|""|""|""|""|""|""|""|""|
the original code contained something like this

when "0010"|"0100"|"1000"|"0001"|"1111"=> 

just way more cases

jeremiah-c-leary:
Hey @joerg-de ,

I again apologize for the late response.

I had to re-write my parser to address other issues I was having which will appear in the 3.0.0 release.  This pull request was part of the decision to do the re-write.

The new parser can handle the situation you pointed out.

Therefore, I am going to close this pull request.

Regards,

--Jeremy

Issue #350, opened 2020-03-18
Built In Configurations for backwards compatibility


jeremiah-c-leary:
The changes proposed for release `2.0.0` affect how rules are named and some of their default values.  In order to make `2.0.0` a more seemless transition, and provide a means for users to manage their own configurations, the concept of **build in configurations (BIC)** has been proposed (see issue #343 for some initial concepts). 

This issue includes the following tasks:

[x] Decide on a name for this feature.  Is "Built In Configurations" good enough?
[x] Determine where the configurations will be stored in the directory hierarchy
[x] Define format of the configurations
[ ] Determine how to let users know these configurations exist 
[ ] Define initial set of configurations to support
[x] Determine how to test the configurations with unit tests
[ ] How can users submit their own configurations for inclusion with VSG
[ ] How can users include their local configurations as an option
[ ] Should users be able to disable existing built in configurations?

m-kru:
@jeremiah-c-leary I have an idea how we can provide it to the user. First I thought we should just prepare example `config.yaml` files and let users edit them. Now I have better idea. We should embed these styles into the vsg. For example user would call vsg in following way: `vsg --style style_name -f file.vhd`. This approach has 2 pros:

1. If you want to use one of the embedded styles you do not need to store `config.yaml` in your repository.
2. If you want to slightly modify the default style, you need to only define an 'overlay' for embedded style. This makes `config.yaml` which you need to store much shorter.

jeremiah-c-leary:
@m-kru That makes a lot of sense.  So VSG could provide a series of styles to choose from, and the user could pick which one they want.  If there are modifications, then they just make a configuration and pass it via the  `-c` option.

```bash
 $vsg --style minimal -c test_config.yaml
 $vsg --style legacy -c my_config.yaml
```
In those cases, the style would be the base configuration and the `-c` would modify it.

Issue #393, opened 2020-05-09
library_001 reports wrong alignment for context (VHDL-2008)


staerz:
**Environment**
VSG 1.6.1., Fedora

**Describe the bug**
The library_001 complains in a context environment (VHDL-2008) where the indentation is actually correct.

**To Reproduce**
Run VSG on the following piece of code:

```
context components is
  library fpga;
  context fpga.interfaces;
end context;
```
**Expected behavior**
I would not expect library_001 to complain here as the indentation is correct.

**Additional context**
Pun intended :rofl:

See here: https://www.doulos.com/knowhow/vhdl_designers_guide/vhdl_2008/vhdl_200x_merged/#context

jeremiah-c-leary:
I think I need to add some context rules also.

staerz:
> I think I need to add some context rules also.

Yes, I think that would be helpful :+1: 

(FYI: I'm just testing VSG 1.12.0 and that one, of course, still remains.)

jeremiah-c-leary:
Hey @staerz ,

Could you checkout the issue-393 branch and let me know if that fixes the issue for you.

I added 27 rules for context declarations.  I have not implemented context_028 yet, which will align comments within a context declaration.  I wanted your feedback before I tackled that one.

staerz:
Hi @jeremiah-c-leary,

I have cloned the repo, checked out the branch and ran the install script, but it seems it doesn't bring a difference.

How can I check if the version in that branch is really being used? Of course, I tried `vsg --version`, but this only gives me a "VHDL Style Guide (VSG) version 2.0.0", but not the commit SHA.

The reason I'm asking is that I don't see any change.

The file I'm checking is this:

```vhdl
context constants is
  library IEEE;
  context IEEE.IEEE_STD_CONTEXT;
  library PoC;
  use PoC.utils.all;
  library fpga;
  use fpga.fpga_cst.all;
end context;
```

and then later (in a different file, using this context) this:
```vhdl
library fpga;
--! @cond
  context fpga.constants;
--! @endcond

package fpga_if is
```

jeremiah-c-leary:
hmm...I would expect your first file to look like:

```vhdl
context constants is

  library IEEE;
  context IEEE.IEEE_STD_CONTEXT;
  library PoC;
  use PoC.utils.all;
  library fpga;
  use fpga.fpga_cst.all;

end context constants;
```

[...]

Issue #378, opened 2020-04-11
There are no rules for package body.


m-kru:
We miss rules for `package body`.

jeremiah-c-leary:
Agree.  It would seem they should be fairly easy to implement.  Similar to the architecture rules.
I am going to assign this to release 2.1.0 so we can focus on the main goal of 2.0.0

jeremiah-c-leary:
Well @m-kru ,

It has been awhile, but I have just added rules for package bodies.

I am going to leave this open for now, but will close it when I release 3.0.0 if I do not hear back from you.

Take care,

--Jeremy

Issue #290, opened 2019-11-21
Multiple signal definitions with the same class and type.


m-kru:
I would like to be able to define multiple signals with the same class and type as follows:
```
signal a,
       b,
       c : std_logic;
```

Is there any rules configuration that enables this and checks for the proper alignment?

Writing multiple signals definitions in this way break 3 rules:
```
  signal_015                |        186 | multiple signals defined in a single declaration
  signal_016                |        186 | ensure signal declaration is on a single line.
  instantiation_024         |        189 | Rewrite positional assignment to explicit assignment
```

Breaking `instantiation_024` indicates, that there is a bug at the vhdl file parsing stage.

jeremiah-c-leary:
hmm...I just added signal_015 and signal_016, so I am not surprised those fired.  However, instantiation_024 should not have flagged anything.  This will have to be investigated.

As for allowing the above formatting, are you wanting to enforce the formatting or just not flag it if it occurs?

For example:

Violation
``` vhdl
  signal a, b, c : std_logic;

  signal x : std_logic;
  signal y : std_logic;
  signal z : std_logic;

```
Fix
```vhdl
  signal a,
            b,
            c : std_logic;

  signal x,
            y,
            z : std_logic;
```

We could take signal_016 as a starting point to enforce that type of formatting.

As a side note, that is in interesting way to define signals.  Each signal is on it's own line, which I like, but there is only one type for all the signals.  I would have to see a file fully defined that way, but I suspect it would be easier to read.

m-kru:
I want to enforce this style for multiple signal definitions. I want vsg to fire an error when it encounters something like this:
```vhdl
signal a, b, c : std_logic_vector(7 downto 0);
```

jeremiah-c-leary:
so...signal_015 will fire if a configurable number of signals are defined over any number of lines.

You would like a rule that fires if a configurable number of signals are defined over a single line.

Violation
```vhdl
  signal a, b, c : std_logic_vector(7 downto 0);
```
Fix
```vhdl
  signal a,
            b,
            c : std_logic_vector(7 downto 0);
```

m-kru:
I took a look at these rules and I am a bit overwhelmed by how rules that work on multiple lines are implemented. I would need considerable amount of time to implement what you have written. I also do not like the fact, that if 2 rules contradicts, than user still needs to manually disable one of them. 

[...]

PR #397, opened 2020-05-17
Implementation of Issue 350

jeremiah-c-leary:
This pull request implements the built in styles from issue #350


docs/styles.rst
```diff
@@ -0,0 +1,15 @@
...
+They can be accessed by using the **--style** command line option.
+
++---------------+--------------------------------------+
+| Configuration | Description                          |
++---------------+--------------------------------------+
+| legacy        | Rules based on the pre 2.0.0 release |
+| minimal       | Only indenting is checked            |
++---------------+--------------------------------------+
+
+These default configurations provide a base which can be further modified using the **--configuration** option.
```
m-kru, line 14: This suggests, that default configurations are not complete/finished, which will not be true. Their main goal is to use them, modifying/extending them is just an extra feature.

vsg/__main__.py
```diff
@@ -33,6 +33,7 @@ def parse_command_line_arguments():
     parser.add_argument('-b', '--backup', default=False, action='store_true', help='Creates a copy of input file for comparison with fixed version.')
     parser.add_argument('-oc', '--output_configuration', default=None, action='store', help='Write configuration to file name.')
     parser.add_argument('-rc', '--rule_configuration', default=None, action='store', help='Display configuration of a rule')
+    parser.add_argument('--style', action='store', choices=get_installed_styles(), help='Uses built in style')
```
m-kru, line 36: Here https://github.com/jeremiah-c-leary/vhdl-style-guide/issues/350 `built in configuration` name is used. In the code and documentation `style`. I propose to use one term, `Built-in Styles`. Of course the argument stays as `--style`, `--built-in-style` would be too long.

vsg/__main__.py
```diff
@@ -42,6 +43,42 @@ def parse_command_line_arguments():
         return parser.parse_args()
 
 
+def get_installed_styles():
```
m-kru, line 46: `get_built_in_styles` / `get_builtin_styles` ?

vsg/__main__.py
```diff
@@ -42,6 +43,42 @@ def parse_command_line_arguments():
...
+    lStyles = os.listdir(sStylePath)
+    for sStyle in lStyles:
+        if sStyle.endswith('.yaml'):
+            with open(os.path.join(sStylePath, sStyle)) as yaml_file:
+                tempConfiguration = yaml.full_load(yaml_file)
+            lReturn.append(tempConfiguration['name'])
+    return lReturn
+
+
+def read_style(commandLineArguments):
```
m-kru, line 65: `read_built_in_style` / `read_builtin_style`?

vsg/__main__.py
```diff
@@ -76,9 +113,9 @@ def validate_file_exists(sFilename, sConfigName):
         sys.exit(1)
 
 
-def read_configuration_files(commandLineArguments):
+def read_configuration_files(dStyle, commandLineArguments):
```
m-kru, line 116: dStyle can still be `None` here, right?

vsg/__main__.py
jeremiah-c-leary: I eventually went with `--style` because `--built-in-configuration` was too long.

Maybe instead of `Uses built in style` it should be `Use predefined style` or `Use available style`?

vsg/__main__.py
jeremiah-c-leary: That makes sense.

vsg/__main__.py
jeremiah-c-leary: It would be an empty dictionary:
```python
dStyle = {}
```

jeremiah-c-leary:
>When it comes to tests. I think in tests directory we should create test_styles / style_tests subdirectory. This directory would contain common *.vhd files that would serve as inputs to all style tests. In the styles subdirectories we would have proper test_style_xxx.py and fixed *.vhd files for given style.

I think that is a great idea.  I will update accordingly.

docs/styles.rst
jeremiah-c-leary: I will elaborate this sentence.

docs/styles.rst
```diff
@@ -0,0 +1,15 @@
...
+======
+
+VSG supports several predefined styles.
+They can be accessed by using the **--style** command line option.
+
++---------------+--------------------------------------+
+| Configuration | Description                          |
++---------------+--------------------------------------+
+| legacy        | Rules based on the pre 2.0.0 release |
+| minimal       | Only indenting is checked            |
```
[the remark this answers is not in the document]
jeremiah-c-leary: So before this feature I would have resisted defaulting to a minimum style because creating configurations can be time consuming.  However, with these styles it makes it easier to users to pick a style that closely matches their own, or even pick one from a list of styles.

If we switch to a minimal style we could add a note to the end of a run that states something like:
```mono
NOTE:  Consider starting with a built in style using the --style argument.
   legacy    : rules based on the pre 2.0.0 release
   lowercase : vhdl keywords, signal names etc... are lowercase
```
... if `--style` or `--configuration` was not used.  Something to nudge the user in the right direction.

So instead of VSG providing a single opinion on what your code should look like and forcing the user to modify that opinion, VSG would provide no opinion and allow the user to pick one from those provided.  That selection is a starting point that can be further modified with the `--configuration` option.

docs/styles.rst
```diff
@@ -0,0 +1,15 @@
+Styles
+======
+
+VSG supports several predefined styles.
+They can be accessed by using the **--style** command line option.
```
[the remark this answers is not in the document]
jeremiah-c-leary: How about 'enabled'?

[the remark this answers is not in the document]
m-kru: Might be confusing as we use enable/disable terms for rules, and styles are not rules.

docs/styles.rst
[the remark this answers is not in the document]
m-kru: The problem with the defaults, as we have seen, is that it is hard to maintain backward compatibility in some cases. The default style (no `--style` or `--configuration`) probably shouldn't be allowed. Maybe we should immediately return and print info to the user that he should use `--style`, `--configuration` or both parameters? Or maybe `style` should become positional argument?

docs/styles.rst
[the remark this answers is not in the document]
jeremiah-c-leary: utilized?

vsg/__main__.py
m-kru: I like predefined.

vsg/__main__.py
m-kru: In case of change to predefined, it would be `get_predefined_styles`.

vsg/__main__.py
m-kru: Ok, I just wanted to make sure, that you can call this without any style, just configuration.

docs/styles.rst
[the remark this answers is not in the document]
jeremiah-c-leary: hmmm....making `--style` a required argument is an interesting idea.  It would force the user to investigate what current styles are available.  Probably best not do define a default value either.

docs/styles.rst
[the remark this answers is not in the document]
m-kru: We have come to the point where I do not know. I am not a native speaker, and I am not sure which one would be better `used` or `utilized`.

docs/styles.rst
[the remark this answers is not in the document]
m-kru: On the other hand, someone might to define his own style, then forcing him to use --style parameter does not seem like a good solution.

docs/styles.rst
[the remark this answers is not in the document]
jeremiah-c-leary: lol...I am a native speaker and it isn't clear to me either.  I think I will go with used.  I updated the entire `style` documentation.  Hopefully it will be clearer.

docs/styles.rst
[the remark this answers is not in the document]
jeremiah-c-leary: Maybe there is a `none` style and that is the default?

If I was running this for the first time, I would want VSG to do something or least tell me what I need to do to get it to do something.

docs/styles.rst
```diff
@@ -0,0 +1,128 @@
...
+They can be used with the **--style** command line option.
+
+The table below lists the built in styles available
+
++---------------+--------------------------------------------+
+| Style         | Description                                |
++---------------+--------------------------------------------+
+| legacy        | Rules based on the pre 2.0.0 release       |
+| lowercase     | All keywords and identifiers are lowercase |
++---------------+--------------------------------------------+
```
m-kru, line 14: This is just example right? What I mean is that each style would need separate section. It is not enough to write 1 sentence. The question arises, how we should name styles. lowercase is not good. I have an idea. Maybe you try to describe only legacy style, as I guess this is the one you use. Also probably do not name it legacy, but write in the doc that it is the style that was used as the default one for vsg so far.

docs/styles.rst
```diff
@@ -0,0 +1,128 @@
...
+
+* Structure
+
+  * No single line sequential statements using the when keyword
+  * No code after the case when statements
+  * Split if/elsif/else/end if into separate lines
+  * Removing comments from instantiation and component ports and generics
+  * No more than two signals can be declared on a single line
+
+lowercase
```
m-kru, line 56: Do you use this lowercase style? If no, then probably there is no point in adding styles just to have more styles.

docs/styles.rst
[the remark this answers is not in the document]
m-kru: We might enforce the user to provide one of:
1. `--style`
2. `--configuration`
3. `--style` and `--configuration`

If user provides neither `--style` nor `--configuration` he will get an error.

docs/styles.rst
jeremiah-c-leary: that was more of an example of what could be done.  

At one point we switched VSG from solely my style to one that is more...hmm...I can't find the word...less my style?.  So the new style became the default.  It would seem we should name it something and then I can name my style something besides legacy.  You could even add your own style.  The way you want to format signal declarations is really interesting and I never would have thought of it.

maybe this feeds back into the other discussion on whether to make `--style` a positional argument?

docs/styles.rst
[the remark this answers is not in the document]
jeremiah-c-leary: If we keep `--style` optional, then the default behavior will be however the rules are currently defined.   I think that is a reasonable position to take.  It addresses my "VSG should do something" issue.

Then that raises a question of why even have this feature.  I could always just use the `--configuration` option.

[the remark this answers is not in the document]
m-kru: Ok, I have one idea. Style becomes positional argument. This way we enforce user to do something to get to know about styles. We provide one special style named `minimal` or `base` with indentation only.

There is still one thing in my mind. What if user wants to define his own style?
We can tell him to use minimal/base style and apply configuration. But it requires him to call vsg in following way `vsg minimal/base --configuraiton my_config`, quite long, on the other hand he can always define an alias.
We can also accept PR with styles. I do not think there will be many.

docs/styles.rst
jeremiah-c-leary: correct, this is just an example to illustrate the concept.

docs/styles.rst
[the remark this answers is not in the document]
jeremiah-c-leary: k, I think I am going to go back on something I said before:
> So instead of VSG providing a single opinion on what your code should look like and forcing the user to modify that opinion, VSG would provide no opinion and allow the user to pick one from those provided.

VSG should have an opinion on what makes well formatted VHDL code.  That opinion is it's base rule set out of the box.

There are over 350 rules and for the legacy style I modified 26 of them (I will probably need to modify more).  They were all capitalization and alignment rules.  Those 26 represent just 7.4% of the total number of rules.  I want to believe that means the majority of the rules would be agreeable to most people.

The `--style` is there to provide the user with other possibilities.

It is almost like VSG should say...

"I believe this is the most readable formatted VHDL code.  If you do not agree, then I have a couple of other styles that people like.  If you do not like any of those, then you can make make up your own style."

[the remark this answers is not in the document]
m-kru: Do I understand correctly, the legacy style will be the default one?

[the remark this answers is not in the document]
jeremiah-c-leary: The current rule base, with the lower casing, would be the default.  Plus anything new that gets added, like #393, #378, etc...  As those new rule sets get implemented, then the other styles would be updated.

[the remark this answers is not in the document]
m-kru: I have one more idea. Lets make the legacy style the default. We disable all the rules in the code (no rule is enabled in any .py file). When `--style` is not passed to the vsg, then vsg chooses the legacy style as the default one. What do you think?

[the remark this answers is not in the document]
jeremiah-c-leary: This might be the point at which we need to test out what I have implemented.  I just pushed a a version to this branch with three styles:  base, disabled and legacy.  I set the base to be the default.

I ran a little experiment on the `spi_slave.vhd` file and switched back and forth between the base and legacy styles.  It is pretty cool to be able to switch between them quickly.

How about you checkout this branch and give it try.  It might help solidify an approach.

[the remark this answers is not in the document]
m-kru: Ok, but tomorrow. I got to go right now.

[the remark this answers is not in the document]
m-kru: `python setup.py install --user` does not install styles, so vsg does not work.
```
FileNotFoundError: [Errno 2] No such file or directory: '/home/mkru/.local/lib/python3.8/site-packages/vsg-1.2.0-py3.8.egg/vsg/styles'
```

I think `styles` directory should be placed next to `vsg` directory, not inside. `vsg` should contain Python sources related to vsg.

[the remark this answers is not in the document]
jeremiah-c-leary: > `python setup.py install --user` does not install styles, so vsg does not work.

That is frustrating.  It must have something to do with the `find_packages` function in the `setup.py` file.

> I think styles directory should be placed next to vsg directory, not inside. vsg should contain Python sources related to vsg.

Will setup.py include them in the install?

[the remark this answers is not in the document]
m-kru: We have to investigate this, but I have seen projects, where directories sitting next to the directory with actual Python code are installed, so people somehow achieve this.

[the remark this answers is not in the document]
jeremiah-c-leary: @m-kru I figured out the issue.  I did not include an `__init__.py` file in the `vsg/styles` directory.  The function `find_packages` will only include directories with that file.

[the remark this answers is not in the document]
m-kru: I now copies styles directory, however without styles .yml files.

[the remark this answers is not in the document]
jeremiah-c-leary: I think I got it this time.  Please try again.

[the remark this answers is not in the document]
m-kru: Ok, works.
So final questions.
1. What we do in case when user does not provide `--style` argument, do we use some style as the default one, or do we inform him, that he must provide `--style` argument? If we use some default style, then I propose the legacy style to be the default one. We can even name this style default.
2.  Do we provide any style with minimal number of rules enabled? 

In my humble opinion disabled style makes no sense.

Btw. I thnk the note at the end misses `the` word.
```
NOTE: Refer to THE online documentation at https://vhdl-style-guide.readthedocs.io/en/latest/index.html for more information.
```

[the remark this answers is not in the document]
jeremiah-c-leary: > What we do in case when user does not provide `--sytle` argument...
We would default to a style, which is currently `base`.
> If we use some default style, then I propose the legacy style...
The legacy style has all of my capitalization and alignment formatting.  That is just an artifact of where I learned to write VHDL.  I think VSG should focus on what provides the most readable VHDL, which I think is currently the base rules.

I should rename `legacy` to something else, but I am having trouble thinking up a name.

> Do we provide any style with minimal number of rules enabled?

I think we should try.  I am a little worried about Phase 1 though:
image

Phase 1 we could disable certain rules, maybe all of them?
Phase 2 we could skip (disable all rules in phase 2)
Phase 3 we could skip (disable all rules in phase 3)
Phase 4 we would do
Phases 5 through 7 we would skip (disable all rules in those phases)

I currently do not have a method to skip phases.  Setting up a configuration to disable the rules in the phase would require constant maintenance.  I could put a skip phases or execute phases parameter in the style YAML file:

```yaml
skip_phases:
    - 1
    - 2
    - 3
    - 5
    - 6
    - 7
```
Pass that list to the `rule_list` object and let it go.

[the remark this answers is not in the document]
m-kru: I guess some rules from phases 2-7 may fail, if the structure is weird. I would opt for enabling phase 1 and 4 in the minimal/base style. But I do not know if all rules should be enabled I would have to read them first.

You still have not answered if `--style` will be mandatory?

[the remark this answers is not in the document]
jeremiah-c-leary: > You still have not answered if --style will be mandatory?

I don't think it should be mandatory.  We should default to the base rule set and you can override it with another style.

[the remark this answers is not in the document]
jeremiah-c-leary: well, I implemented the `skip_phase` and ran just phase 4 (Indent) on the PIC.vhd file and it seemed to work.  The file is well structured though.  I will try other files.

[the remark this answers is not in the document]
jeremiah-c-leary: I ran against every file in the code_examples and it actually did fairly well.  There are a couple of issues.
  1)  There are some alignment rules that should probably be ran in the indent phase:
          process_020
          concurrent_003
  2)  Ending parenthesis in generic and port statements when they are not on their own line have indent issues.

Other than that it looked pretty good.  I will work on committing the changes.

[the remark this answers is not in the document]
m-kru: 1. Trying to run alignment rules in indent phase probably means that something is not implemented in proper way.
2. This is somehow related to the 1.

I think we should think for  a while on how we handle indentation and how we handle alignment in case of multi line issues, and what is what. Right now I wonder how would I handle indentation/alignment in the signal declaration approach I would like to apply. https://github.com/jeremiah-c-leary/vhdl-style-guide/issues/290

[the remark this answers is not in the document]
jeremiah-c-leary: > I think we should think for a while on how we handle indentation and how we handle alignment in case of multi line issues, and what is what.

That is a good point.

```mono
I      a <= b or
A         c and d when
A         e = 1 else
A         a;
```
Where I = Indent and A = alignment.
```mono
I  CLK_PROC : process (a ,b
A                      c, d) is
I   begin
I   end process CLK_PROC;
```

```mono
I   case state
I     when idle =>
I       n_state <= write;
I       n_write_en <= a or b or
A                     c and d;
I    when write =>
```

It would seem that it a line starts a VHDL element? that it would be an indent.  If a line continues an element then it would be an alignment.  This distinction only applies to the beginning of the line.

```mono
I  entity FIFO is
I  port (
I   I_WR_EN : in std_logic;
I   I_RD_EN : in std_logic
I  );
I  end entity FIFO;
```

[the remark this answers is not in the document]
jeremiah-c-leary: @m-kru I just pushed the minimal style and tests.  You can see some examples of alignments not occuring in some of the files, due to the alignment phase being skipped.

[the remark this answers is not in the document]
m-kru: Hmm, I like your idea about marking lines with I and A. However right now it does not happen when the file is parsed, does it?

vsg/styles/minimal.yaml
```diff
@@ -0,0 +1,9 @@
+# The name must be a single word
+name : minimal
+skip_phase:
+  - 1
+  - 2
+  - 3
+  - 5
+  - 6
+  - 7
```
m-kru, line 9: This can be written in following way `skip_phase: [1, 2, 3, 5, 6, 7]`

vsg/tests/comment/test_rule_comment.py
```diff
@@ -27,7 +27,7 @@ def test_rule_010(self):
 
     def test_rule_010_library(self):
         oRuleList = rule_list.rule_list(oFileLibrary)
-        oRuleList.check_rules()
+        oRuleList.check_rules([])
```
m-kru, line 30: Maybe there should be default value for this argument, there will be no need to pass empty table `[]`.

docs/styles.rst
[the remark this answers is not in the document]
jeremiah-c-leary: The lines are not marked `I` and `A` specifically, but it can be inferred from the `line.indentLevel` attribute:

```mono
2     case state
3       when idle =>
4         n_state <= write;
4         n_write_en <= a or b or
None                    c and d;
3       when write =>
```

vsg/styles/minimal.yaml
jeremiah-c-leary: thx, I will make the change.

vsg/tests/comment/test_rule_comment.py
jeremiah-c-leary: That makes perfect sense.  I will update.

jeremiah-c-leary: I just pushed an update.

vsg/styles/minimal.yaml
jeremiah-c-leary: I just pushed an update.

docs/styles.rst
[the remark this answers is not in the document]
jeremiah-c-leary: > Right now I wonder how would I handle indentation/alignment in the signal declaration approach I would like to apply. #290

In this case the indent/alignment would be:

```mono
2      signal write_en,
None          read_en,
None          full,
None          almost_full : std_logic;
```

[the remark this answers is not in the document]
jeremiah-c-leary: hey @m-kru just wanted to ping you about this PR.  Do you think this is done enough to merge into master?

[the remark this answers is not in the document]
m-kru: I will analyze it again today.

docs/styles.rst
```diff
@@ -0,0 +1,137 @@
...
+
+VSG supports several predefined styles.
+They can be used with the **--style** command line option.
+
+The table below lists the built in styles available
+
++---------------+--------------------------------------------+
+| Style         | Description                                |
++---------------+--------------------------------------------+
+| base          | Current rule set                           |
```
m-kru, line 12: What does it mean current rule set? If we can not guarantee backward compatibility for `base` style, then we should probably remove this style.

docs/styles.rst
```diff
@@ -0,0 +1,137 @@
...
+| disabled      | All rules are disabled                     |
```
m-kru, line 13: I would remove `disabled` style. It is super easy to disable all rules in the yaml config file, so user can do it.

docs/styles.rst
```diff
@@ -0,0 +1,137 @@
...
+       indentSize: 3
+     entity_004:
+       case: upper
+   ...
+
+Then use the style and configuration together:
+
+.. code-block:: mono
+
+   $ vsg --style lowercase --configuration my_config.yaml -f fifo.vhd
```
m-kru, line 135: There is no `lowercase` style. Right now I would leave only `legacy` style. There is no reason to artificially add styles. I will add style that I use at some point. We can also accept pull requests from users, but adding styles just to have more styles does not make sense to me.

vsg/__main__.py
```diff
@@ -33,6 +33,8 @@ def parse_command_line_arguments():
     parser.add_argument('-b', '--backup', default=False, action='store_true', help='Creates a copy of input file for comparison with fixed version.')
     parser.add_argument('-oc', '--output_configuration', default=None, action='store', help='Write configuration to file name.')
     parser.add_argument('-rc', '--rule_configuration', default=None, action='store', help='Display configuration of a rule')
+    parser.add_argument('--style', action='store', default='base', choices=get_predefined_styles(), help='Use predefined style')
+#    parser.add_argument('--skip_phase', nargs='+', default=None, help='Skips entire phases of analysis')
```
m-kru, line 37: What about this line? Shoud it be removed or uncommented?

vsg/__main__.py
```diff
@@ -42,6 +44,42 @@ def parse_command_line_arguments():
...
+    lStyles = os.listdir(sStylePath)
+    for sStyle in lStyles:
+        if sStyle.endswith('.yaml'):
+            with open(os.path.join(sStylePath, sStyle)) as yaml_file:
+                tempConfiguration = yaml.full_load(yaml_file)
+            lReturn.append(tempConfiguration['name'])
+    return lReturn
+
+
+def read_predefined_style(commandLineArguments):
```
m-kru, line 66: The parameter of this function should rather be `name`, not `commandLineArguments`.
Having to all command line arguments indicates that something is probably wrong with the logic.

vsg/__main__.py
```diff
@@ -244,6 +286,12 @@ def display_rule_configuration(commandLineArguments, configuration):
         sys.exit(fExitStatus)
 
 
+def validate_files_exist_to_analyze(commandLineArguments):
```
m-kru, line 289: Again, I am not sure if passing `commandLineArguments` is good idea. The actual argument is filename.

vsg/styles/minimal.yaml
```diff
@@ -0,0 +1,4 @@
+# The name must be a single word
+name : minimal
```
m-kru, line 2: `minimal` style is not mentioned in the documentation. Thee are `base`, `disabled`, `legacy`. I would remove `base` and `disabled`, and release only `minimal` and `legacy` in the 2.0.0. release. Maybe it would be good to rename `minimal` to `indent`/`indent_only`?

docs/styles.rst
jeremiah-c-leary: base would be the rule set for 2.0.0 and forward.  The style itself is an empty dictionary so it will pickup what the current rules are set to.  Being an empty set, it would evolve as rules are added.

docs/styles.rst
jeremiah-c-leary: I suppose you could argue that `disabled` is not really a style.  I can remove it.

docs/styles.rst
jeremiah-c-leary: I have updated and replaced lowercase with legacy.

vsg/__main__.py
jeremiah-c-leary: I removed it.

vsg/__main__.py
jeremiah-c-leary: I see what you are saying.  It came about because I was calling `open_configuration_file`, which takes commandLineArguments.  I will work on improving the function.

vsg/__main__.py
jeremiah-c-leary: I agree.

vsg/styles/minimal.yaml
jeremiah-c-leary: I think `indent_only` makes sense.  I still not sure about removing `base`.  I will add documentation for `indent_only`.

docs/styles.rst
jeremiah-c-leary: Maybe `default` instead of `base`?

vsg/__main__.py
jeremiah-c-leary: Update has been pushed.

vsg/__main__.py
jeremiah-c-leary: Update has been pushed.

docs/styles.rst
jeremiah-c-leary: Update has been pushed.

docs/styles.rst
jeremiah-c-leary: Update has been pushed.

vsg/__main__.py
jeremiah-c-leary: I have pushed an update for this comment.

docs/styles.rst
m-kru: Sorry, I really do not understand what is the point of this style. 

> Being an empty set, it would evolve as rules are added.

Isn't this against the idea of builtin styles?
I thought we are going to disable all rules in the python code, and only enable them in styles and configuration files.

vsg/__main__.py
```diff
@@ -63,41 +62,40 @@ def get_predefined_styles():
...
-    if commandLineArguments.style:
-        sFileName = os.path.join(os.path.dirname(__file__), 'styles', commandLineArguments.style + '.yaml')
-        dReturn = open_configuration_file(sFileName, commandLineArguments)
+    sFileName = os.path.join(os.path.dirname(__file__), 'styles', sStyleName + '.yaml')
+    open_configuration_file(sFileName)
     return dReturn
     
 
-def open_configuration_file(sFileName, commandLineArguments):
+def open_configuration_file(sFileName, sJUnitFileName=None):
```
m-kru, line 81: Could you explain the name `sJUnitFileName`? What is its relation with `JUnit`?

docs/styles.rst
jeremiah-c-leary: I think I view the rules currently enforced as a built in style.  Given that, I wanted to let the user know what it was trying to accomplish.

Maybe the question is:  what is value in explicitly calling out the `default` style?  Why have an option that does not change any behavior?

This command:
```bash
$ vsg -f entity.vhd
```
is equivalent to:
```bash
$ vsg -f entity.vhd --style default
```
and looking at it that way, it does not make any sense to include default as a style.

vsg/__main__.py
jeremiah-c-leary: so `sJUnitFileName` is the file name passed to VSG via the `--junit` argument.

It is passed to the  `open_configuration_file` in case there is an error reading the configuration file.  If an error is detected and the `--junit` option was chosen, then the error is pushed to a JUnit XML file via the `write_invalid_configuration_junit_file` function.

m-kru: And what is JUnit per se? What is the idea behind writing such errors to JUnit XML?

docs/styles.rst
m-kru: I think we can make the legacy style the default one. What I mean, is that default should be just an alias to another style. The other solution is to make the minimal/indentation style the default one. Maybe such approach would be even better.

vsg/__main__.py
jeremiah-c-leary: oh, JUnit?  As I understand it, JUnit is a Java unit test platform.  They came up with an XML format for showing errors in your Java testsuite.

Bamboo, the Continuous Integration (CI) tool we use at my work, had a parser for JUnit XML files.  Bamboo would parse these files and then indicate if there was a failure.

I wrote log parsers for all the tools we used:  Quartus, Spyglass, Questasim, Questa CDC etc... and generated these JUnit files.  It gave us finer control over whether we failed a job in Bamboo.

For example, I ran VSG on a simple entity file with two violations and I get this printed to the screen:
```mono
python bin/vsg -f debug/entity.vhd --junit blah.xml
File:  debug/entity.vhd
Phase 3 of 7... Reporting
Total Rules Checked: 198
Total Violations:    2
----------------------------+------------+--------------------------------------
  Rule                      |  line(s)   | Solution
----------------------------+------------+--------------------------------------
  whitespace_012            |          3 | Remove all but 1 blank lines below.
  entity_016                |          5 | Remove blank line(s) above "end entity" keywords.
----------------------------+------------+--------------------------------------
```
If I add `--junit vsg.xml`, I also get this file:
```xml
<?xml version="1.0" ?>
<testsuite errors="0" hostname="Rimmer" failures="2" timestamp="2020-06-08T12:45:08" tests="1" time="0" name="vhdl-style-guide">
  <properties>
  </properties>
  <testcase name="debug/entity.vhd" time="0" classname="failure">
    <failure type="Failure">
      entity_016: 5 : Remove blank line(s) above &quot;end entity&quot; keywords.
      whitespace_012: 3 : Remove all but 1 blank lines below.
    </failure>
    <failure type="Failure">
      entity_016: 5 : Remove blank line(s) above &quot;end entity&quot; keywords.
      whitespace_012: 3 : Remove all but 1 blank lines below.
    </failure>
  </testcase>
  <system-out>
  </system-out>
  <system-err>
  </system-err>
</testsuite>

hmm...the data seems to be repeated.  I will have to fix that.

Anyway...the CI server will pull in the XML file and fail the run if it finds any failures.  You can then navigate to the errors and the data in the XML file will be displayed. 
```

m-kru: Ok, I understand. Is your log parsing tool open source?

jeremiah-c-leary: Those log parsers are not open source.  However, I did start a new side project  https://github.com/jeremiah-c-leary/eda-log-file-warning-suppressor that essentially performs the same function.  I currently only supports Mentor Graphics Precision, Microsemi Designer and Xilinx Vivado log files.

Maybe you could git a try and let me know what you think.

jeremiah-c-leary:
Hey @m-kru,
There are a lot conversations on this pull request, and I think they  are all resolved.  Would you object to me merging this into master?

Let me know if there is anything we still need to resolve.

m-kru:
@jeremiah-c-leary I am not sure about the added styles `base`, `indent_only`, `legacy`. I do not think we need `base`, and I would rename `legacy`.

jeremiah-c-leary:
> I am not sure about the added styles base, indent_only, legacy. I do not think we need base, and I would rename legacy.

I am okay with removing `base`.  I am having a hard time figuring out an alternate name for `legacy`.

What  do you think of:
  `1.0_style`
  `pre_2.0`
  `authors`
  `learys`
  `jcls`
  `jeremiahs`

...or we just do not have a `legacy`.

jeremiah-c-leary:
So thinking about this a little more...what would you name your coding style?

m-kru:
I have not thought about this yet. With one word there is no way to make it descriptive. Maybe we can think of some convention, for example name of stars, digital elements or famous engineers, just loudly thinking :). You need description of each style anyway.

jeremiah-c-leary:
@m-kru ,
I removed the `base` style and renamed `legacy` to `jcl`.

let me know if there are any other changes.

m-kru:
What about making all the rules disabled by default and just enabling them in styles, config files?
Do we agree on this, or do you prefer it the way it has been so far?
Or maybe you agree, but there should be new PR for this?

jeremiah-c-leary:
I think keeping the current method, where VSG prefers a code style and the user modifies that style, is better a model.  I think most users would keep the default rules.  The obvious exceptions would be case and alignment rules.

In issue #425, @staerz mention changing 38 rules out of 350+.  So that is about 11% of the rules.  So @staerz agrees with 89% of the current rule set.

staerz:
I agree, make VSG come with the default set of rules and a default style.

I think VSG is very well documented already - and if you simply add the doc for the styles, everybody will be able to understand and switch to a different style.

But having rules disabled or styles hiding some rules by default is a bad idea (imho).

jeremiah-c-leary:
I am going to merge this into master and we can continue the discussion on another thread if we want.

nfrancque:
This feature isn't working on master.  The configuration isn't taking effect.

If I change line 79 of __main__.py from 
` open_configuration_file(sFileName)`

to 

`dReturn = open_configuration_file(sFileName)`

it seems to work again.  Otherwise, I think the file is being read but the configuration is not being returned anywhere.

Issue #424, opened 2020-06-14
Prefix rule for process and generate labels


marph91:
**Is your feature request related to a problem? Please describe.**
I would like to define prefixes for the process and generate labels.

**Describe the solution you'd like**
A rule for process, respectively generate, to define the prefixes.

m-kru:
Definitely feasible. It would be quite easy to implement, maybe you could try to make PR :)?

jeremiah-c-leary:
Interesting idea.  You could use an existing prefix rule as a template.  It would just be a matter of pulling the label and checking if it starts with the prefix.

marph91:
Sure, I will take a look. For my designs, the prefixes are mostly `PROC_` respectively `GEN_`. Should I use them as default or do you prefer something else?

m-kru:
They are fine as the defaults, but remember that by default everything should be lowercase, so it would be `proc_` and `gen_`.

jeremiah-c-leary:
Should this rule be on by default?

Also, I just committed some tokenizing code when I fixed a bug for issue #427.  It might make your checks easier.

m-kru:
This is something, that we already tried to discuss in some other thread. In my opinion all rules should be disabled by default, and enabling would be done in builtin styles and configuration files.

marph91:
@m-kru: Are you sure about lowercase? In the [docs](https://vhdl-style-guide.readthedocs.io/en/latest/process_rules.html#process-017) its written uppercase. Also when I tried it, the process labels were converted to uppercase.

@jeremiah-c-leary: I think it should be disabled, to keep it consistent to the other prefix and suffix rules.

m-kru:
@marph91 Yes, I am sure. This doc probably needs an update. We are leaning towards making everything lowercase by default, and then uppercase in style and configuration files. It is more consistent approach.

marph91:
That makes sense. Now I see that the case is checked by a separate rule.

Btw.: The [shebang of the vsg executable](https://github.com/jeremiah-c-leary/vhdl-style-guide/blob/3e987c7a3981473156a665dfb710ce337df86bd6/bin/vsg#L1), invoked on my system (Opensuse tumbleweed) Python 2, which led to strange errors. I had to fix it by writing explicitly `python3`. Should I include it in the PR? Not sure if it would break on other distros.

m-kru:
Do not include it in PR, Python 2 is no longer supported.

[...]

PR #436, opened 2020-06-25
Added rules for generate and process label prefixes

marph91:
**Description**
This PR fixes https://github.com/jeremiah-c-leary/vhdl-style-guide/issues/424.

### DIFF
--- a/vsg/rules/generate/rule_017.py
+++ b/vsg/rules/generate/rule_017.py
@@ -11,7 +11,7 @@ class rule_017(prefix_rule):
     def __init__(self):
         prefix_rule.__init__(self, 'generate', '017', 'isGenerateLabel')
         self.prefixes = ['gen_']
-        self.solution = 'Generate'
+        self.solution = 'Generate label'
 
     def _extract(self, oLine):
-        return utils.extract_class_identifier_list(oLine)
+        return utils.extract_label(oLine)
--- a/vsg/rules/generate/rule_017.py
+++ b/vsg/rules/generate/rule_017.py
@@ -1,0 +1,17 @@
+
+from vsg.rules import prefix_rule
+from vsg import utils
+
+
+class rule_017(prefix_rule):
+    '''
+    Generate rule 017 checks for prefixes in generate statement labels.
+    '''
+
+    def __init__(self):
+        prefix_rule.__init__(self, 'generate', '017', 'isGenerateLabel')
+        self.prefixes = ['gen_']
+        self.solution = 'Generate'
+
+    def _extract(self, oLine):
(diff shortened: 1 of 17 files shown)


vsg/rules/generate/rule_017.py
```diff
@@ -0,0 +1,17 @@
...
+    Generate rule 017 checks for prefixes in generate statement labels.
+    '''
+
+    def __init__(self):
+        prefix_rule.__init__(self, 'generate', '017', 'isGenerateLabel')
+        self.prefixes = ['gen_']
+        self.solution = 'Generate'
+
+    def _extract(self, oLine):
+        return utils.extract_class_identifier_list(oLine)
```
jeremiah-c-leary, line 17: Could you try the `utils.extract(label)` function?
It seems more appropriate in this instance.
This would apply to process_036 also.

vsg/rules/process/rule_036.py
```diff
@@ -0,0 +1,16 @@
...
+
+class rule_036(prefix_rule):
+    '''
+    Process rule 036 checks for prefixes in process labels.
+    '''
+
+    def __init__(self):
+        prefix_rule.__init__(self, 'process', '036', 'isProcessLabel')
+        self.prefixes = ['proc_']
+        self.solution = 'Process'
```
jeremiah-c-leary, line 13: So I noticed the `_get_solution` method in the `prefix_rule` class shows the following:

```python
    def _get_solution(self, iLineNumber):
        return self.solution + ' identifiers should start with one of ' + str(self.prefixes) + ' prefixes.'
```
With `self.solution` equal to `Process`, the solution printed out would be:
```mono
Process identifiers should start with one of proc_ prefixes
```
Which seems misleading.
If you could make the following updates:
  1)  Change `self.solution` to `Process label`
  2)  Remove `identifiers` from the above code in `prefix_rule.py`
  3)  Find the other rules using `prefix_rule` and add `identifiers` to their `self.solution` attribute

I think that should clarify what the user should fix.

vsg/rules/generate/rule_017.py
marph91: Sure, that makes sense.

vsg/rules/process/rule_036.py
marph91: I adapted each solution of a prefix rule to the corresponding docstring. The solutions should be more accurate now.

marph91:
Hi Jeremy,

I fixed the requested remarks. Feel free to correct me if there are more.

jeremiah-c-leary:
Hey @marph91 ,

It looks great.  I will merge it in.

PR #437, opened 2020-07-01
Updates for issue #290:

jeremiah-c-leary:
Initial implementation of analysis portion of rule signal_017:

  1)  Added documentation
  2)  Added test
  3)  Added analysis portion of rule signal_017

@m-kru ,
This is the analysis portion of a rule to address issue #290.  Let me know what you think and we can talk about how to do the fix.

### DIFF
--- a/docs/signal_rules.rst
+++ b/docs/signal_rules.rst
@@ -346,3 +346,27 @@ This rule checks the signal declaration is on a single line.
    signal sig1 : std_logic;
 
    signal sig2 : std_logic;
+
+signal_017
+==========
+
+This rule handles formatting of signal declarations with more than one signal.
+
+**Violation**
+
+.. code-block:: vhdl
+
+   signal sig1, sig2
+     sig3, sig4,
+     sig5
+     : std_logic;
+
+**Fix**
+
+.. code-block:: vhdl
+
+   signal sig1,
+          sig2,
+          sig3,
+          sig4,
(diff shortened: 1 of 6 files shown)


m-kru:
@jeremiah-c-leary does the new rule 017 conflict with any other rules? By conflict I mean, if there are any 2 rules, that when enabled together lead to contradiction.

jeremiah-c-leary:
rules `signal_015` and `signal_016` would conflict.  I was planning on making rule `signal_017` a superset of those rules and then depricate `signal_015` and `signal_016`.  Essentially, there would be a single rule to handle formatting of signal declarations.

There is still a lot of work to be done for `signal_017`, but I was wondering if this is what you are looking for.  Then I was going to  work on consolidating the other rules into `signal_017`.

m-kru:
From user perspective, this is what I was looking for. Note that the are more classes where such rule could be applied. Namely `constant`, `variable`, `file`. I think it would be good idea to create some base class.

jeremiah-c-leary:
hmm...you are probably correct about creating a base clase.  At the moment I want to get this rule working and then I will refactor it into a base clase.

jeremiah-c-leary:
Hey @m-kru ,

I am going to close this pull request.  This implementation is now longer relevant with the parser re-write.

Hope things are going well for you.

--Jeremy

Issue #402, opened 2020-06-01
Warning or Severity option for rules


ericjmcd:
I would like to have a feature where we could flag a rule with warnings only.  i.e. If the rule is violated, a warning message (or violation message) is still printed but the return code is still zero.  I'm using VSG with Jenkins and pre-commit hooks so I would like to allow some rule violations to occur with warnings without causing the check/build to fail.

Ideally, I'd like there to be a severity that is configurable per rule so that users could tailor the importance of each.  Maybe something similar to: violation, warning, format.

I'm trying to get my team to adopt VSG and I don't want to force them to drastically modify their code to pass or to disable a bunch of rules that we actually want to enforce.

jeremiah-c-leary:
Hey @ericjmcd 
I think such a thing could be done.  There are some implications to be worked through.

I am curious about this statement:
> I'm trying to get my team to adopt VSG and I don't want to force them to drastically modify their code to pass or to disable a bunch of rules that we actually want to enforce.

I would assume your team would use the `--fix` option to bring their code into compliance.

Could you elaborate on this a little?

ericjmcd:
You're right - I forgot that the --fix option was there (although I added it for my "make reformat" target).  I was thinking it was unreliable due to to the "generic name with 'port'" bug that I submitted (and you quickly fixed - thanks!) but I haven't had any problems since.  Once we get our config file to match the style that we like, I think we'll apply --fix liberally to get our projects in line.

PS - Thanks for making such a great tool.  If I wasn't already overcomitted to too many other things, I'd love to contribute.  Maybe someday I will.

jeremiah-c-leary:
> PS - Thanks for making such a great tool. If I wasn't already overcomitted to too many other things, I'd love to contribute. Maybe someday I will.

Thanks for the kind words, and I am glad you are finding it useful.

How soon would you like to see this implemented?  I am trying to figure out priorities between several tickets.

ericjmcd:
It's not a major hurry for us - just something that will make the tool more useful so we can soften the pass/fail criteria for our CI/CD pipeline (I could pipe the output to file and parse it and then report the warnings - I wish jUnit supported warnings but I don't believe it does).
For implementation, what about:
* add a global "severity" with either "error" or "warning" (defaulting to "error")
* allow rules to change the global "severity" to "error" or "warning"
* if a rule fails with "warning" severity, don't record it as a failure but still output it to stdout/sterr but change the output:
[...]

Issue #417, opened 2020-06-07
Add JSON output format for violations


jeremiah-c-leary:
To support integration with editors, outputting the violations in JSON format makes transferring data between applications easier.

This is in support of issue #401

qarlosalberto:
I'm testing the new features and changing my code. I get this json output:

```
{
  "colibri/test/sample.vhd": {
    "violations": [
      {
        "rule": "architecture_024",
        "linenumber": 28,
        "solution": "Add architecture name keyword."
      },
      {
        "rule": "entity_019",
        "linenumber": 15,
        "solution": "Add the entity name."
      },
      {
        "rule": "process_012",
        "linenumber": 22,
        "solution": "Add \"is\" keyword after the closing parenthesis."
      },
      {
        "rule": "process_016",
        "linenumber": 22,
        "solution": "Add a label for the process."
      },
      {
        "rule": "process_018",
        "linenumber": 25,
        "solution": "Add a label for the \"end process\"."
      }
    ]
  }
}
```

What is the the purpose of ** "colibri/test/sample.vhd"** ? What you do you think about?:

```
{
  "files" : 
  [
    {
      "file_path" : "/path/to/file",
      "violations": [
        {
          "rule": "architecture_024",
          "linenumber": 28,
          "solution": "Add architecture name keyword."
        },
        {
          "rule": "entity_019",
          "linenumber": 15,
          "solution": "Add the entity name."
        }
      ]
    },
    {
      "file_path" : "/path/to/file2",
      "violations": [
        {
          "rule": "architecture_024",
          "linenumber": 28,
          "solution": "Add architecture name keyword."
        },
        {
          "rule": "entity_019",
          "linenumber": 15,
          "solution": "Add the entity name."
        }
      ]
    }
  ]
}
```

jeremiah-c-leary:
I can update to that format.

jeremiah-c-leary:
I just pushed an update to match the above JSON format.  Could you update and double check?

Thanks,

--Jeremy

qarlosalberto:
Yes, its fine!

jeremiah-c-leary:
Awesome, I will close this issue.

PR #439, opened 2020-07-04
Issue 402

jeremiah-c-leary:
This is the implementation of issue #402 .

Am I doing something wrong?  I have an issue and a pull request.  This seems redundant.


docs/rule_severity.rst
```diff
@@ -0,0 +1,68 @@
...
+           "length_001":{
+               "severity": "Warning"
+           }
+       }
+   }
+
+Defining Severity Levels
+------------------------
+
+VSG supports defini
```
ericjmcd, line 38: Looks like you forgot to finish the comments here.

ericjmcd:
I've tested this a little bit and it seems really good.  However, the junit output still counts all the findings are failures.  E.g. I set the default to a Guideline as warning and overrode 1 rule to be a Warning and the stdout matches:
```
architecture_026          | Guideline  |         48 | Align colon with right most colon.
variable_assignment_005   | Guideline  |        158 | Inconsistent alignment of ":=" in group of lines.
concurrent_005            | Warning    |        361 | Remove label on concurrent assignment.
```
However, the junit output has all failures.
```
<failure type="Failure">
      architecture_026: 48 : Align colon with right most colon.
...
```
I think extract_junit_testcase() needs to be updated?  I didn't want to look too deep in the code.  I tried to find a good spec on the JUnit XML but couldn't find anything concrete. This is the best I found:
[JUnit+Format](https://help.catchsoftware.com/display/ET/JUnit+Format)

There doesn't seem to be any support for warnings so I think for severities that aren't error/failures they should just be ignored in the JUnit output.

jeremiah-c-leary:
> There doesn't seem to be any support for warnings so I think for severities that aren't error/failures they should just be ignored in the JUnit output.

What you would think about separating the warnings into the different severities in the JUnit XML file.  The thinking being you could have the CI tool ignore the severities you do not care about.  Although, that seems like extra work for the user.  So maybe just removing them from the JUnit XML file would be best.

docs/rule_severity.rst
jeremiah-c-leary: agree, not sure what happened there.  Probably distracted by something shiny.

jeremiah-c-leary:
@ericjmcd ,
I added the filtering of warnings from the JUnit file.  Could you update and check if it is working as you expect it work?

Thanks,

--Jeremy

ericjmcd:
The test case classname is "failure" which is a bit odd (maybe change that just to "vsg"?) but there aren't any <failure> nodes now.  However, the cumulative failures at the <testsuite> level still shows the errors.

For what it's worth, I don't think I will end up using the unit output as the Jenkins plugins (xUnit) and the JUnit format don't support the severity construct so the errors alone aren't terribly valuable.  I think I'll write my own stdout parser so I can track/plot errors/warnings/guidelines/etc.  Of course, if time permits...

jeremiah-c-leary:
Take a look at issue #417 .  I could catagorize by severity, or add the severity to each violation.

jeremiah-c-leary:
> The test case classname is "failure" which is a bit odd (maybe change that just to "vsg"?) but there aren't any <failure> nodes now. However, the cumulative failures at the <testsuite> level still shows the failures (e.g. 26 files with Guideliens will show failures=26)..

I am a little unclear about what the issue is.  Could you elaborate or post some example JUnit files?

jeremiah-c-leary:
Hey @ericjmcd ,

Curious if this is ready to merged or if there is anything else I need to do.

Thanks,

--Jeremy

Issue #504, opened 2020-12-23
Adjust behaviour for comments (allow_comments configuration)


staerz:
**Is your feature request related to a problem? Please describe.**
It is rather an inconsistency on how rules apply, but has likely a personal bias.

**Describe the solution you'd like**
A few rules request a blank line above and have a configuration option `allow_comments`.
This configuration option, as the name suggests, allows that the line before is a comment and not a blank line.

This option effectively is like disabling the rule entirely as then a blank line above is not checked anymore.

Instead of the current behaviour for this option, the first line above the comment which is not a comment should be checked and this should be required to be a blank line.

For example, `entity_003` has this option.
```vhdl
library xyz;

-- A comment describing my_entity
-- And maybe we need some more lines to explain more details.
entity my_entity is
```
This example has an blank line (line 2). In the current behaviour of VSG this blank line can also be removed if the option `allow_comments` is set and VSG won't complain.

Instead, with the option `allow_comments` set, this line 2 should be enforced to be a blank line.

**Describe alternatives you've considered**
There is none :wink:

jeremiah-c-leary:
Hey @staerz ,

I have an initial attempt at this feature.  When you get a chance, could you check it out and let me know if it meets your needs?

Thanks,

--Jeremy

staerz:
Hi @jeremiah-c-leary,

I updated my local VSG installation to 0bec7656 and ran it over my code base again but couldn't find a difference.

I have tried to violate `architecture_003` and `entity_003`, both have `allow_comments` configured to be true, but I see no effect.

Could you be more specific what I should try?

The particular piece of code I ran over was this one:
```vhdl
--! @cond
library fpga;
  context fpga.interfaces;

--! @endcond
--! Non-locking multiplexer for incoming AVMM master interfaces.
entity avmm_master_mux is
  port (
    --! Input clock
    clk_avmm_i     : in    std_logic;
    --! Reset (sync to clock)
    rst_avmm_i     : in    std_logic;
    --! Incoming AVMM master interfaces
    avmm_masters_i : in    t_avmm_masters;
    --! Outgoing AVMM master interface
    avmm_master_o  : out   t_avmm_master
  );
end entity avmm_master_mux;

--! Clocked implementation of non-locking multiplexer for incoming AVMM master interfaces.
architecture behavioral of avmm_master_mux is
```
vs. this without the blank lines.

Both give the very same result.

jeremiah-c-leary:
[...]

PR #523, opened 2021-01-16
Proposed implementation of issue #504

jeremiah-c-leary:
Configuration documentation has been updated for proposed changes to line above rule.

### DIFF
--- a/docs/configuring_previous_line_rules.rst
+++ b/docs/configuring_previous_line_rules.rst
@@ -4,24 +4,21 @@ Configuring Previous Line Rules
 There are rules which will check the contents on lines above code structures.
 These rules allow enforcement of comments and blank lines.
 
-There are several options to these rules, but they fall into two categories:  allow and require.
-Require options will check the feature exists above the line in question.
-Allow options provide exceptions to the require options.
+There are several options to these rules, which can be selected by using the :code:`method` option:
 
-+---------------------+---------+---------+--------------------------------------------------+
-| Option              | Values  | Default | Description                                      |
-+=====================+=========+=========+==================================================+
-| require_blank_line  | Boolean |  True   | Requires a blank line on the line above.         |
-+---------------------+---------+---------+--------------------------------------------------+
-| allow_comment       | Boolean |  False  | Allows comments in the line above.               |
-+---------------------+---------+---------+--------------------------------------------------+
-| require_comment     | Boolean |  False  | Requires a comment on the line above.            |
-+---------------------+---------+---------+--------------------------------------------------+
-| blank_above_comment | Boolean |  False  | Enforces a blank line above comments.            |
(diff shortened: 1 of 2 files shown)


docs/configuring_previous_line_rules.rst
```diff
@@ -0,0 +1,133 @@
...
++=====================+==================================================+
+| require_blank_line  | Requires a blank line on the line above.         |
++---------------------+--------------------------------------------------+
+| no_code             | Either a blank or comment on the line above.     |
++---------------------+--------------------------------------------------+
+| allow_comment       | Either a blank or comment on line above, with a  |
+|                     | blank line above the comment.                    |
++---------------------+--------------------------------------------------+
+| require_comment     | Requires a comment on the line above, with a     |
+|                     | blank line above the comment.                    |
```
staerz, line 20: Sorry, first time giving multi line suggestions on github - I think I missed the header and last line ...
```suggestion
| require_blank_line  | Requires a blank line on the line above.                   |
+---------------------+------------------------------------------------------------+
| no_code             | Either a blank line; or comment(s) on the line(s) above.   |
+---------------------+------------------------------------------------------------+
| allow_comment       | Either a blank line; or comment(s) on line(s) above and a  |
|                     | blank line above the comment(s).                           |
+---------------------+------------------------------------------------------------+
| require_comment     | Comment(s) required on the line(s) above and a             |
|                     | blank line above the comment(s).                           |
```

docs/configuring_previous_line_rules.rst
```diff
@@ -0,0 +1,133 @@
...
+   rule :
+     entity_003:
+        method : require_blank_line
+
+.. NOTE:: All examples below are using the rule **entity_004**.
+
+Example: require_blank_line
+###########################
+
+The following code would fail with this option:
```
staerz, line 36: I think you should always give the fail example first, then the pass.

This is in alignment with the general rule violations: you first always show the violation and then the fix.

staerz: hm ... ok, the difference here is that it's not a fix that you suggest, right? ...

Well, in that case, simply make the examples have pass and fail in the same order :wink:

staerz:
Well, you came up with a very good name, I think `method` is very good!

I'd say the documentation is good like now (modulo my few suggestions modulo rearranging the table after I messed up the suggestion :rofl: )

docs/configuring_previous_line_rules.rst
jeremiah-c-leary: I thought about making it a fix, but thought just illustrating what would trigger and what would not trigger the rule would be  clearer.

jeremiah-c-leary:
> I'd say the documentation is good like now

Awesome...I will start working on the changes.

PR #579, opened 2021-05-01
Issue 576

imd1:
Work complete on issue 576. Unit tests added and pass.


vsg/tests/generic/test_rule_600.py
```diff
@@ -0,0 +1,41 @@
...
+        self.assertEqual(oRule.identifier, '600')
+
+        lExpected = [14, 15]
+
+        oRule.analyze(self.oFile)
+        self.assertEqual(lExpected, utils.extract_violation_lines_from_violation_object(oRule.violations))
+
+    def test_rule_600_w_uppercase(self):
+        oRule = generic.rule_600()
+        oRule.suffixes = ['_G']
```
jeremiah-c-leary, line 33: Could you make this '_W' and update the lExpected?

jeremiah-c-leary:
Hey @imd1 ,

It looks really good.  I just had the one comment.

--Jeremy

vsg/rules/process/rule_600.py
```diff
@@ -0,0 +1,19 @@
...
+
+class rule_600(token_suffix):
+    '''
+    Constant rule 600 checks for suffixes in process labels.
+    '''
+
+    def __init__(self):
+        token_suffix.__init__(self, 'process', '600', lTokens)
+        self.suffixes = ['_proc']
+        self.solution = 'Process labels'
```
imd1, line 19: I was going to ask you exactly what these lines do in each rule, as I'm not confident they are consistent across the new rules I added, and some of the existing rule .e.g some do not have `solution`

imd1: Maybe you can review these lines in each rule to double check?

jeremiah-c-leary: I reviewed the `token_suffix.py` file and `self.solution` is not used.  I was at one point, but line 51 generates the solution based on the token being suffixed and the possible suffixes defined the by rule.  I must not have cleaned up the other rules.

jeremiah-c-leary: The suffix rule uses `self.suffixes` in its checks.

vsg/tests/generic/test_rule_600.py
imd1: i have renamed the test...

jeremiah-c-leary:
Are we ready to merge this?

imd1:
> Are we ready to merge this?

Yes

PR #606, opened 2021-06-11
Issue 598

imd1:
@jeremiah-c-leary : I've added some more vertical alignment conditions - hopefully the docs are self explanatory!

### DIFF
--- a/docs/configuring_keyword_alignment.rst
+++ b/docs/configuring_keyword_alignment.rst
@@ -166,7 +166,7 @@ Rule Specific Keyword Alignment Configuration
           data_o         : in std_logic
       );
 
-#. :code:`if_control_statements_end_group` - if set to :code:`True` any line with if control statement ends the group of lines that should be aligned and starts new group.
+#. :code:`if_control_statements_ends_group` - if set to :code:`True` any line with if control statement ends the group of lines that should be aligned and starts new group.
    By default set to :code:`True`.
 
     **Violation**
@@ -181,7 +181,7 @@ Rule Specific Keyword Alignment Configuration
           hold_transmission <= '1';
       end if;
 
-    **Fix (if_control_statements_end_group = True)**
+    **Fix (if_control_statements_ends_group = True)**
 
     .. code-block:: vhdl
 
@@ -193,7 +193,7 @@ Rule Specific Keyword Alignment Configuration
           hold_transmission <= '1';
       end if;
 
-    **Fix (if_control_statements_end_group = False)**
+    **Fix (if_control_statements_ends_group = False)**
 
     .. code-block:: vhdl
 
@@ -205,13 +205,14 @@ Rule Specific Keyword Alignment Configuration
           hold_transmission <= '1';
       end if;
 
-#. :code:`case_control_statements_end_group` - if set to :code:`True` any line with case control statement ends the group of lines that should be aligned and starts new group.
+#. :code:`case_control_statements_ends_group` - if set to :code:`True` any line with case control statement ends the group of lines that should be aligned and starts new group.
    By default set to :code:`True`.
 
     **Violation**
 
     .. code-block:: vhdl
 
+      data_valid_before    <= '1';
       case A is
           when A =>
               X <= F;
@@ -224,11 +225,13 @@ Rule Specific Keyword Alignment Configuration
           when others =>
             null;
       end case
+      data_valid_after       <= '1';
 
-    **Fix (case_control_statements_end_group = True)**
+    **Fix (case_control_statements_ends_group = True)**
 
     .. code-block:: vhdl
 
+      data_valid_before <= '1';
       case A is
           when A =>
               X   <= F;
@@ -241,11 +244,54 @@ Rule Specific Keyword Alignment Configuration
           when others =>
               null;
       end case
+      data_valid_after <= '1';
 
-    **Fix (case_control_statements_end_group = False)**
+    **Fix (case_control_statements_ends_group = False)**
 
     .. code-block:: vhdl
 
+      data_valid_before <= '1';
+      case A is
+          when A =>
+              X         <= F;
+              XY        <= G;
+              XYZ       <= H;
+          when B =>
+              a         <= I;
+              ab        <= h;
+              c         <= a;
+          when others =>
+              null;
+      end case
+      data_valid_after  <= '1';
+
+#. :code:`case_keyword_statements_ends_group` - if set to :code:`True` and :code:`case_control_statements_ends_group` is set to :code:`False`, any line with case when statement ends the group of lines that should be aligned and starts new group. If set to :code:`False`, this configuration has no effect and :code:`case_control_statements_ends_group` rules are followed. 
+   By default set to :code:`False`.
+
+    **Violation**
+
+    .. code-block:: vhdl
+
+      data_valid_before    <= '1';
+      case A is
+          when A =>
+              X <= F;
+              XY <= G;
+              XYZ <= H;
+          when B =>
+              a <= I;
+              ab <= h;
+              c <= a;
+          when others =>
+            null;
+      end case
+      data_valid_after       <= '1';
+
+    **Fix (case_keyword_statements_ends_group = True and case_control_statements_ends_group = False)**
+
+    .. code-block:: vhdl
+
+      data_valid_before <= '1';
       case A is
           when A =>
               X   <= F;
@@ -258,6 +304,63 @@ Rule Specific Keyword Alignment Configuration
           when others =>
               null;
       end case
+      data_valid_after <= '1';
+
+    **Fix (case_keyword_statements_ends_group = False and case_control_statements_ends_group = False)**
+
+    .. code-block:: vhdl
+
+      data_valid_before <= '1';
+      case A is
+          when A =>
+              X         <= F;
+              XY        <= G;
+              XYZ       <= H;
+          when B =>
+              a         <= I;
+              ab        <= h;
+              c         <= a;
+          when others =>
+              null;
+      end case
+      data_valid_after  <= '1';
+
+
+#. :code:`loop_control_statements_ends_group` - if set to :code:`True` any line with loop control statement (including for and while loops) ends the group of lines that should be aligned and starts new group.
+   By default set to :code:`False`.
+
+    **Violation**
+
+    .. code-block:: vhdl
+
+      data_valid_before    <= '1';
+      for index in 4 to 23 loop
+          data_valid <= '0';
+          hold_transmission <= '1';
+      end loop;
+      data_valid_after       <= '1';
+
+    **Fix (loop_control_statements_ends_group = True)**
+
+    .. code-block:: vhdl
+
+      data_valid_before <= '1';
+      for index in 4 to 23 loop
+          data_valid        <= '0';
+          hold_transmission <= '1';
+      end loop;
+      data_valid_after <= '1';
+
+    **Fix (loop_control_statements_ends_group = False)**
+
+    .. code-block:: vhdl
+
+      data_valid_before     <= '1';
+      for index in 4 to 23 loop
+          data_valid        <= '0';
+          hold_transmission <= '1';
+      end loop;
+      data_valid_after      <= '1';
 
 .. NOTE:: If given keyword alignment rule has any of the above keyword alignment specific configuration, then it is explicitly noted in the documentation of this rule.
 
--- a/docs/configuring_keyword_alignment.rst
+++ b/docs/configuring_keyword_alignment.rst
@@ -265,7 +265,7 @@ Rule Specific Keyword Alignment Configuration
       end case
       data_valid_after  <= '1';
 
-#. :code:`case_keyword_statements_ends_group` - if set to :code:`True` and :code:`case_control_statements_ends_group` is set to :code:`False`, any line with case when statement ends the group of lines that should be aligned and starts new group. If set to :code:`False`, this configuration has no effect and :code:`case_control_statements_ends_group` rules are followed. 
+#. :code:`case_keyword_statements_ends_group` - if set to :code:`True` and :code:`case_control_statements_ends_group` is set to :code:`False`, any line with :code:`case` or :code:`end case` statements ends the group of lines that should be aligned and starts new group. If set to :code:`False`, this configuration has no effect and :code:`case_control_statements_ends_group` rules are followed. 
    By default set to :code:`False`.
 
     **Violation**
--- a/docs/configuring_keyword_alignment.rst
+++ b/docs/configuring_keyword_alignment.rst
@@ -205,7 +205,7 @@ Rule Specific Keyword Alignment Configuration
           hold_transmission <= '1';
       end if;
 
-#. :code:`case_control_statements_ends_group` - if set to :code:`True` any line with case control statement ends the group of lines that should be aligned and starts new group.
+#. :code:`case_control_statements_ends_group` - if set to :code:`True`, any line with case control statements (:code:`case`, :code:`when` or :code:`end case`) ends the group of lines that should be aligned and starts new group. If set to :code:`False`, no line with case control statements ends the group of lines that should be aligned and starts a group. If set to :code:`break_on_case_or_end_case`, any line with :code:`case` or :code:`end case` ends the group of lines that should be aligned and starts new group.
    By default set to :code:`True`.
 
     **Violation**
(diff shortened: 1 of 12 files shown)


docs/configuring_keyword_alignment.rst
```diff
@@ -241,11 +244,54 @@ Rule Specific Keyword Alignment Configuration
...
+          when B =>
+              a         <= I;
+              ab        <= h;
+              c         <= a;
+          when others =>
+              null;
+      end case
+      data_valid_after  <= '1';
+
+#. :code:`case_keyword_statements_ends_group` - if set to :code:`True` and :code:`case_control_statements_ends_group` is set to :code:`False`, any line with case when statement ends the group of lines that should be aligned and starts new group. If set to :code:`False`, this configuration has no effect and :code:`case_control_statements_ends_group` rules are followed.
```
jeremiah-c-leary, line 268: So if I understand this correctly, if `case_control_statements_ends_group` is set to `True` it would stop aligning when it hit the `case` keyword.  However, you could not control alignment across the `when` keywords.

When `case_control_statements_ends_group` is set to `False` then alignment would occur through the case statement.  Essentially aligning everything under all the `when` statements with everything outside the `case` statement.

With the new `case_keyword_statements_ends_group` set to `True` then the statements under the `when` keywords will not be aligned.

I feel like I missed something.

docs/sequential_rules.rst
```diff
@@ -100,8 +100,10 @@ This rule checks the alignment of the **<=** operators over consecutive sequenti
 
 Following extra configurations are supported:
 
-* :code:`if_control_statements_end_group`,
-* :code:`case_control_statements_end_group`.
+* :code:`if_control_statements_ends_group`,
```
jeremiah-c-leary, line 103: The only problem I see with changing `end` with `ends` is the people who already have it configured and using `end`.  VSG should report an error when attempting to configure a rule with an invalid option.  However, is there a better way to communicate this?

We could put it in the release notes, but I doubt many people would read it.

I did add a rule deprication feature so VSG will flag rules that are removed or split up.  Would it be worth it to add option deprication?  If a configuration attempts to change an option that used to exist, `if_control_statements_end_group`, we could report something like:

ERROR:  Rule sequential_005 option if_control_statements_end_group.
   This option has been changed to if_control_statements_ends_group.

imd1: I believe this is just a documentation issue i.e. the docs have `end`  but the code is correct and uses `ends`

docs/configuring_keyword_alignment.rst
imd1: The idea of the new `case_keyword` attribute is:
* if left at a default of False, the existing behavior of the `case_control` attribute is maintained for backwards compatibility
* this existing behavior is:
   * True means the `case/end case` AND* `when` keywords ends a group (*important)
   * False means that NEITHER `case/end case` NOR `when` ends a group
 * so the only new behavior happens when `case_keyword` is True and `case_control` is False. This means that `case/end case` ends a group (from the `case_keyword`) but `when` does NOT end a group (from the `case_control`)
 * the new attribute allows `when` statements to be configured differently to `case/end case` statements
 * I added some assignments above and below the case statements to try and show what happens outside the case statements

Maybe take another look at the docs for this, and feel free to improve the explanation - I found it hard!

docs/sequential_rules.rst
jeremiah-c-leary: I just checked and you are correct.

docs/configuring_keyword_alignment.rst
jeremiah-c-leary: k, I think I understand what you were going for.  Would you be up for either of these changes:

Combine options
============

  1)  remove `case_keyword_statements_ends_group`
  2)  add third option to `case_control_statements_ends_group`:  `True`, `False`, `break_at_when`

Change new option
=============

  1)  rename `case_keyword_statements_ends_group` to `case_when_keyword_ends_group`
  2)  This would perform the same effect, but would only take affect if `case_control_statements_ends_group` is `True`

I think that might clear up the documentation, because `case_control_statements_ends_group` would not be counter intuitive.

imd1: Before I look at your suggestions, I've just realised that the docs were wrong and I've made a correction. Please could you take a look at the new docs and then let me know how you want to proceed - happy to continue with your options once I'm sure you've accounted for the doc error I've corrected....

imd1: I've looked at your options and I think there is still a slight misunderstanding...

In the combine options change, the third option would have to be something like ends_at_case_and_case_end.

I cannot see the advantage of the `change new option` idea as it stands

jeremiah-c-leary: I pulled down your pull request and I am working on an example so I can understand how the options work.

So given this example code:
```vhdl
  1
  2 architecture rtl of fifo is
  3
  4 begin
  5
  6   label : process () is
  7   begin
  8
  9     a <= 1;
 10
 11     b_blah <= 2;
 12
 13     case blah is
 14
 15       when 0000 =>
 16
 17         a <= 1;
 18         b_blah <= 2;
 19
 20       when 0010 =>
 21
 22         a <= 2;
 23         b_blah2 <= 3;
 24
 25     end case;
 26
 27     a <= 1;
 28
 29     b_blah <= 2;
 30
 31   end process label;
 32
 33 end architecture rtl;
```
If I use this configuration:
```yaml
  1 rule:
  2   sequential_005:
  3     blank_line_ends_group: False
  4     comment_line_ends_group: True
  5     if_control_statements_ends_group: True
  6     case_control_statements_ends_group: False
  7     case_keyword_statements_ends_group: True
  8     loop_control_statements_ends_group: False
```
Results in the following:
```vhdl
  1
  2 architecture rtl of fifo is
  3
  4 begin
  5
  6   label : process () is
  7   begin
  8
  9     a      <= 1;
 10
 11     b_blah <= 2;
 12
 13     case blah is
 14
 15       when 0000 =>
 16
 17         a       <= 1;
 18         b_blah  <= 2;
 19
 20       when 0010 =>
 21
 22         a       <= 2;
 23         b_blah2 <= 3;
 24
 25     end case;
 26
 27     a      <= 1;
 28
 29     b_blah <= 2;
 30
 31   end process label;
 32
 33 end architecture rtl;
```
I would have expected the `=>` on lines 17 and 18 to be one column to the left.
With this configuration:
```yaml
  1 rule:
  2   sequential_005:
  3     blank_line_ends_group: False
  4     comment_line_ends_group: True
  5     if_control_statements_ends_group: True
  6     case_control_statements_ends_group: True
  7     case_keyword_statements_ends_group: False
  8     loop_control_statements_ends_group: False
```
results in:
```vhdl
 13     case blah is
 14
 15       when 0000 =>
 16
 17         a      <= 1;
 18         b_blah <= 2;
 19
 20       when 0010 =>
 21
 22         a       <= 2;
 23         b_blah2 <= 3;
 24
 25     end case;
```
Is that backwards from what you were intending?~

imd1: To see the third behavior, try this:

  1 rule:
  2   sequential_005:
  3     blank_line_ends_group: False
  4     comment_line_ends_group: True
  5     if_control_statements_ends_group: True
  6     case_control_statements_ends_group: False
  7     case_keyword_statements_ends_group: False
  8     loop_control_statements_ends_group: False

imd1: Let's number the three examples in order: 1,2 (and 3 if you try my suggestion)

In example 1, the groups end on lines 12 and 25 only

In example 2, the groups end on lines 12, 15, 20 and 25

In example 3, when you've done it, there will be no groups end!

Examples 2 and 3 are existing behavior, example 1 is the new one I want to be able to achieve

imd1: To answer you question, its working as I intended :-)

imd1: It's slightly hard to achieve what I want and keep backwards compatibility, which is essential I believe. I think what I've done works, it's just coming up with a good name etc

jeremiah-c-leary: Example 1
-----------
```yaml
rule:
  sequential_005:
    blank_line_ends_group: False
    case_control_statements_ends_group: False
    case_keyword_statements_ends_group: False
```
```vhdl
 15       when 0000 =>
 16
 17         a       <= 1;
 18         b_blah  <= 2;
 19
 20       when 0010 =>
 21
 22         a       <= 2;
 23         b_blah2 <= 3;
```
Where it is aligned with lines 9, 11, 27 and 29.
Where lines 9, 11, 17, 18, 22, 23, 27 and 29 are aligned.

Example 2
------------
```yaml
rule:
  sequential_005:
    blank_line_ends_group: False
    case_control_statements_ends_group: True
    case_keyword_statements_ends_group: False
```
```vhdl
 15       when 0000 =>
 16
 17         a      <= 1;
 18         b_blah <= 2;
 19
 20       when 0010 =>
 21
 22         a       <= 2;
 23         b_blah2 <= 3;
```
Where lines 9, 11, 27 and 39 are not aligned with 17, 19, 22 and 23.
Where lines 17 and 18 are aligned.
Where lines 22 and 23 are aligned.
Where lines 17 and 18 are not aligned with 22 and 23.

Example 3
------------
```yaml
rule:
  sequential_005:
    blank_line_ends_group: False
    case_control_statements_ends_group: False
    case_keyword_statements_ends_group: True
```
```vhdl
 15       when 0000 =>
 16
 17         a       <= 1;
 18         b_blah  <= 2;
 19
 20       when 0010 =>
 21
 22         a       <= 2;
 23         b_blah2 <= 3;
```
Where lines 17, 18, 22 and 23 are aligned.

okay....I think I get it now.

`case_control_statements_ends_group` will create groups on `case`, `end case`, and `when` keywords.
`case_keyword_statements_ends_group` will create groups on `case` and `end case` keywords only.

Is my understanding aligned with your intent?

docs/configuring_keyword_alignment.rst
```diff
@@ -205,13 +205,14 @@ Rule Specific Keyword Alignment Configuration
           hold_transmission <= '1';
       end if;
 
-#. :code:`case_control_statements_end_group` - if set to :code:`True` any line with case control statement ends the group of lines that should be aligned and starts new group.
+#. :code:`case_control_statements_ends_group` - if set to :code:`True` any line with case control statement ends the group of lines that should be aligned and starts new group.
```
jeremiah-c-leary, line 208: How about his for an update:
... if set to :code:`True` any line with a case control statement (`case`, `when`, `end case`) ends the group of lines that ...

docs/configuring_keyword_alignment.rst
imd1: I believe so
* now you have seen the intent, please suggest better ways of configuring this if you can. I quite like making case_control a 3 way flag
* also the PR code coverage needs looking at...again feel free to suggest improvements as my Python is not expert

jeremiah-c-leary: > now you have seen the intent, please suggest better ways of configuring this if you can. I quite like making case_control a 3 way flag

I will add some comments to the `align_tokens_in_region_between_tokens_unless_between_tokens.py` file.

vsg/rules/align_tokens_in_region_between_tokens_unless_between_tokens.py
```diff
@@ -57,7 +57,10 @@ def __init__(self, name, identifier, lTokens, left_token, right_token, lUnless):
         self.configuration.append('if_control_statements_ends_group')
         self.case_control_statements_ends_group = False
         self.configuration.append('case_control_statements_ends_group')
-
+        self.case_keyword_statements_ends_group = False
```
jeremiah-c-leary, line 60: Remove lines 60 and 61 as we will re-use `case_control_statements_ends_group`

vsg/rules/align_tokens_in_region_between_tokens_unless_between_tokens.py
```diff
@@ -142,6 +145,19 @@ def analyze(self, oFile):
                            dAnalysis = {}
 
                    if self.case_control_statements_ends_group:
```
jeremiah-c-leary, line 147: We need a function to replace the condition in the if statement.  Maybe something like `is_case_control_enabled`.
```python
147                if is_case_control_enabled(self.case_control_statement_ends_group):
```
The function has the single task of determining if the correct values are set in `self.case_control_statement_ends_group`.
Create the function at the end of the file:
```python
def is_case_control_enabled(config):
    if config == 'skip_whens':
        return True
    return config
```
There is probably a better name than `skip_whens`, but we can check for it before returning the config value.  This should allow us to define any number of exceptions to the original behavior while keeping backwards compatibility.

vsg/rules/align_tokens_in_region_between_tokens_unless_between_tokens.py
```diff
@@ -142,6 +145,19 @@ def analyze(self, oFile):
...
+                           for iKey in list(dAnalysis.keys()):
+                               if dAnalysis[iKey]['adjust'] != 0:
+                                   oLineTokens = oFile.get_tokens_from_line(iKey)
+                                   sSolution = 'Move ' + dAnalysis[iKey]['token_value'] + ' ' + str(dAnalysis[iKey]['adjust']) + ' columns'
+                                   oViolation = violation.New(oLineTokens.get_line_number(), oLineTokens, sSolution)
+                                   oViolation.set_action(dAnalysis[iKey])
+                                   self.add_violation(oViolation)
+
+                           dAnalysis = {}
+                   elif self.case_keyword_statements_ends_group:
```
jeremiah-c-leary, line 160: We should be able to delete lines 160 through 172 since we would be removing `self.case_keyword_statements_ends_group`

vsg/rules/align_tokens_in_region_between_tokens_unless_between_tokens.py
```diff
@@ -142,6 +145,19 @@ def analyze(self, oFile):
                            dAnalysis = {}
 
                    if self.case_control_statements_ends_group:
+                       if check_for_case_keywords(iIndex + 1, lTokens) or check_for_when_keywords(iIndex + 1, lTokens):
```
jeremiah-c-leary, line 148: Replace the condition in the if statement at line 148 with another function like `is_case_keyword`:
```python
148            if is_case_keyword(self.case_control_statement_ends_group, iIndex, lTokens):
```
Then create the function at the end of the file:
```python
def is_case_keyword(config, iIndex, lTokens):
    if check_for_case_keywords(iIndex + 1, lTokens):
        return True
    if check_for_when_keywords(iIndex + 1, lTokens):
        if config == 'break_at_whens':
           return True
    return False
```
I think that is about all you need to do.

I would update the test first to add the desired 3 values to the existing `case_control_statements_ends_group` before making any changes to this file.

imd1:
@jeremiah-c-leary I'm struggling with making the attribute have 3 levels when it is currently a boolean (`True`, `False`), without having to change the `True` to a string `"True"` which breaks backwards compatibility? My Python is not good enough for this

docs/configuring_keyword_alignment.rst
imd1: Updated

vsg/rules/align_tokens_in_region_between_tokens_unless_between_tokens.py
imd1: Done

vsg/rules/align_tokens_in_region_between_tokens_unless_between_tokens.py
imd1: Done

vsg/rules/align_tokens_in_region_between_tokens_unless_between_tokens.py
imd1: Done

imd1:
@jeremiah-c-leary I've refactored as you suggested, with some necessary tweak to the logic to get the correct behavior. The attribute can either be boolean or a string, which makes me nervous that this is possible regarding reading the config file?

jeremiah-c-leary:
> @jeremiah-c-leary I've refactored as you suggested, with some necessary tweak to the logic to get the correct behavior. The attribute can either be boolean or a string, which makes me nervous that this is possible regarding reading the config file?

It looks good.  The documentation reads better with the two options combined into one.

Are you referring to the config file when you use `-rc` or `-oc`?

imd1:
> > @jeremiah-c-leary I've refactored as you suggested, with some necessary tweak to the logic to get the correct behavior. The attribute can either be boolean or a string, which makes me nervous that this is possible regarding reading the config file?
> 
> It looks good. The documentation reads better with the two options combined into one.
> 
> Are you referring to the config file when you use `-rc` or `-oc`?

We've converted the `ends_group` parameter from a boolean (`True`, `False`) to a boolean + a third option, a string 'break_on..'.  My Python is not strong. Will the code that reads a configuration file cope with this? I guess I could try running VSG, but is there a unit test somewhere covering this situation?

imd1:
@jeremiah-c-leary : please could you pick this PR again

jeremiah-c-leary:
Hey @imd1,

Checking it out now.

PR #615, opened 2021-10-05
Issue 566

imd1:
Solution to issue 566


jeremiah-c-leary:
What is the thinking behind starting with 410 versus 402 for a rule number?

jeremiah-c-leary:
Should we have similar rules for functions?

imd1:
> Should we have similar rules for functions?

Yes, maybe we should, if they do not already exist. Trying to stick to ticket scope as I'm always time limited and would want as much merged to master as early as possible. I recommend raising a new ticket so it can be worked on when time allows.

jeremiah-c-leary:
> Maybe I thought we would move some existing rules from 0xx to 4xx as part of a future tidy up? Not sure. Happy to start with 402 if you'd prefer

I'm fine with keeping them at their current numbers.  I thought figuring out a numbering scheme would have been easy, but there are nuances I keep running into.

jeremiah-c-leary:
> Yes, maybe we should, if they do not already exist. Trying to stick to ticket scope as I'm always time limited and would want as much merged to master as early as possible. I recommend raising a new ticket so it can be worked on when time allows.

I will create a new ticket for functions.

PR #732, opened 2022-02-10
fix typo in example ('disabled' -> 'disable')

bae-pabr:
**Description**
State a clear description of the problem and your solution.

**Screenshots**
If applicable, add screenshots to help explain your solution.

**Additional context**
Add any other context about the problem here including related issue numbers.

### DIFF
--- a/docs/architecture_rules.rst
+++ b/docs/architecture_rules.rst
@@ -522,7 +522,7 @@ This rule allows the user to restrict what can be used for an architecture name.
 
       rule :
         architecture_025 :
-          disabled : False
+          disable : False
           names :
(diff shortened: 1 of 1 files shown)


jeremiah-c-leary:
Hey @bae-pabr,

Thanks for the pull request.

Could you also update the doc string with the same change in this rule file: 
 `vsg/rules/architecture/rule_025.py`.

I have a test that helps ensures the doc string of a rule matches the restructured text documentation.

Thanks,

--Jeremy

bae-pabr:
Thanks for your hint - I think the test should work now.

PR #786, opened 2022-05-16
Fix broken links to Subprogram Body Rules

jobtijhuis-philips:
Fix broken links to Subprogram Body Rules.

Some rules were moved into the Subprogram Body Rules but the references from the old location to the new location were broken &mdash;this PR fixes that.

**PS: I just noticed an error in the contributing page I will also quickly fix that**


jeremiah-c-leary:
Hey @jobtijhuis-philips,

Thanks for the pull request.  Links have been my Achilles heel for some reason.  Probably because I can't write an automated to verify them.   Hmm.....I wonder if I could write a test to at least validate the html would exist.

I will push this to master.

Thanks,

--Jeremy

PR #828, opened 2022-09-30
workaround: create junit result file even if some vhdl files can not parsed

bae-pabr:
**Description**
I want to run vsg on multiple files (file_list) and create a junit result file, but one of them is not parsable and no junit file is created.
I would like to skip unparsable files in the junit result file.

State a clear description of the problem and your solution:
a vhdl file which can not be parsed ('Error: Unexpected token detected...') creates a None-value for testcase, which would lead to an error when generating junit results ('AttributeError: 'NoneType' object has no attribute 'failures'); workaround: do not add None, so junit result file can still be created.

### DIFF
--- a/vsg/__main__.py
+++ b/vsg/__main__.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 
+from cgi import test
 import sys
 import os
 import json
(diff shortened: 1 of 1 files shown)


jeremiah-c-leary:
Hey @bae-pabr,

> I want to run vsg on multiple files (file_list) and create a junit result file, but one of them is not parsable and no junit file is created.
> I would like to skip unparsable files in the junit result file.

Would you rather have the file skipped on have it reported in a junit file.  I should be able to report the file name along with the message.

--Jeremy

bae-pabr:
> Would you rather have the file skipped on have it reported in a junit file. I should be able to report the file name along with the message.

Hi Jeremy,
reporting filename with message sounds like a great solution (if that is possible).

jeremiah-c-leary:
Hey @bae-pabr,

Would you mind closing this PR and opening an issue?  Then I can start working on a solution.

Thanks,

--Jeremy

bae-pabr:
Issue created (https://github.com/jeremiah-c-leary/vhdl-style-guide/issues/831)

PR #897, opened 2022-12-07
Adds ability to read from standard input

### LINKED ISSUE #896: Ability to read VHDL file from standard input
The ability to be able to read the VHDL file to be checked from standard input would make it easier/faster to be able to use VSG in editors asynchronously/without needing to save to be able to get diagnostics. 

**Describe alternatives you've considered**
If the client/program you want to use to wrap can use temp files, that could work, but that would add unnecessary delay

**Additional context**
Specifically for neovim, it's possible to wrap arbitrary programs with a plugin and provide diagnostics based on the output of the program. I believe this is the purpose the `syntastic` output format exists for this program. I have done this with [null-ls](https://github.com/jose-elias-alvarez/null-ls.nvim)

SethGower:
Allows the user to send the VHDL file to VSG from `stdin` instead of reading files on the disk. Resolves #896


jeremiah-c-leary:
Hey @SethGower,

I apologize for the lateness, but I am finally ready to merge this to master.  Thanks for the pull request.

--Jeremy

PR #989, opened 2023-08-06
Issue#975:  Adding "remove" and "add" options to rule generate_011.

### LINKED ISSUE #975: generate_011 does not have the `action` configuration option
**Environment**
v3.16.0

**Describe the bug**
The rule `generate_011` configures an optional VHDL item, and is part of the structure::optional group. However, it does not have an `action` configuration like the [other optional item rules](https://vhdl-style-guide.readthedocs.io/en/latest/configuring_optional_items.html#configuring-optional-items). This means that, even though `generate_011` is part of this group, it isn't affected by setting this `action` configuration for the group.

**To Reproduce**
1. Run `vsg -rc generate_011`
2. See that output does not include the `action: '[add|remove]'` option

**Expected behavior**
I expect generate_011 to have this configuration item so that it can be configured along with the rest of the group.

**Screenshots**
None

**Additional context**
None

jeremiah-c-leary:
Resolves #975


vsg/rules/generate/rule_011.py
```diff
@@ -126,3 +160,15 @@ def manage_labels(oToken, lLabels):
...
+
+
+def remove_keyword(self):
+    if self.action == 'remove':
+        return True
+    return False
+
+
+def create_violation(oToi, iLineNumber, self):
+    sSolution = self.action.capitalize() + ' ' + self.solution
```
JHertz5, line 172: Because self.solution is "Add generate label", if the action is `remove`, `sSolution` here becomes "Remove Add generate label". Perhaps the solution is to change `self.solution` in `init()` to "generate label"? `

JHertz5 (changes requested):
A suggestion to fix a solution message bug.

vsg/rules/generate/rule_011.py
jeremiah-c-leary: @JHertz5 , you are correct.  setting self.solution to "generate label" fixes the issue.  I pushed an updated to the branch.  Let me know if there are any other issues.

--Jeremy

JHertz5: I can confirm that the update has fixed the issue.

JHertz5 (approved):
Thanks @jeremiah-c-leary, that update has resolved the issue. Everything is looking good now.

PR #1010, opened 2023-09-16
Prevent Python from replacing the existing newline characters.

cschlaak:
Prevent Python from replacing the existing newline characters.
From the [documentation](https://docs.python.org/3/library/functions.html#open): "... If newline is '' or '\n', no translation takes place ..."

**Description**
My code base is based on unix line endings '\n'. When running VSG on Windows every file is rewritten with windows line endings '\r\n'. This is due to pythons implementation of 'open'. The added newline argument prevents python from replacing the existing line endings.

### DIFF
--- a/vsg/apply_rules.py
+++ b/vsg/apply_rules.py
@@ -131,7 +131,7 @@ def apply_rules(commandLineArguments, oConfig, tIndexFileName):
 
 def write_vhdl_file(oVhdlFile):
     try:
-        with open(oVhdlFile.filename, 'w', encoding='utf-8') as oFile:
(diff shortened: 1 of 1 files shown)


alonbl:
We probably need to set up the configuration to enforce a format and not preserve existing and add new ones using a different kind.

jeremiah-c-leary:
Morning @cschlaak ,

Wondering if I could get some feedback on @alonbl 's proposed implementation in #1012 .  I like the idea of using the os default, but being able to override it.

If you do approve, would you be okay with closing this pull request after #1012 has been merged to master?


Thanks,

--Jeremy

PR #1011, opened 2023-09-25
Support vsg as pre-commit hook

alonbl:
**Description**

[pre-commit](https://pre-commit.com) is a tool for running various of format and validation tools. It would be nice if it will be possible to use vsg as pre-commit hook.

**Additional context**

Trivial changes provided as a patch set:
* Support of providing files as positional arguments, most tools assumes that the input files may be used as positional arguments, so let's support both notations.
* Support build from shallow git clone without tags, assign a temporary version 0 to this mode.
* Add pre-commit-hook metadata.

Tested and working for me, please let me know what you think.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,8 @@
+- id: vsg
+  name: fix VHDL style
+  description: VHDL Style Guide
+  entry: vsg --fix --jobs=1
+  language: python
+  types:
+    - file
+  files: \.(vhd|vhdl)$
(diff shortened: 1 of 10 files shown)


jeremiah-c-leary:
Morning @alonbl ,

I feel the documentation for `pre-commit` should be moved from the usage page to it's own page.  I worry people could miss this integration opportunity.  I can't decide exactly where it would go.

So at the top level there are the following sections:
* Continuous Integration Servers
* Editor Integration
* Tool Integration
The first two do not apply as `pre-commit` is neither a CI server or an editor.  The second makes sense as it is very generic, however, `pre-commit` is a specific tool and the options listed under **Tool Integration** do not apply.  I suppose we could split the section into specific tools and then generic tools.  So something like this:
* Tool Integration
   * pre-commit
   * generic
Similar to the **Editor Integration** section:
* Editor Integration
  * VIM
  * NeoVim
  * VSCode

What do you think?

--Jeremy

alonbl:
OK, moved documentation to own page, in most cases people who use pre-commit understand how to use the plugins, I do not think we should be so verbose.

jeremiah-c-leary:
Morning @alonbl ,

Looks good.

--Jeremy

jeremiah-c-leary (approved):

jeremiah-c-leary (approved):

alonbl:
Hi @jeremiah-c-leary,
I added another example for the pre-commit configuration.
Alon

alonbl:
Closed via #1017

PR #1014, opened 2023-10-04
Perform atomic write to files

alonbl:
**Description**
Do not overwrite file as we may corrupt the file if something fails, instead write into temporary location and move the result into the final destination.

### DIFF
--- a/vsg/apply_rules.py
+++ b/vsg/apply_rules.py
@@ -150,12 +150,19 @@ def apply_rules(commandLineArguments, oConfig, tIndexFileName):
 
 
 def write_vhdl_file(oVhdlFile):
+    tmpfile = f"{oVhdlFile.filename}.tmp"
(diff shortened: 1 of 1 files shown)


jeremiah-c-leary:
Morning @alonbl ,

I can not think of any reason not to use a temp file and then copy it over.

So I imagine this pull request will be closed when we close #1017 , correct?

--Jeremy

alonbl:
> Morning @alonbl ,
> 
> I can not think of any reason not to use a temp file and then copy it over.

It is atomic move not copy, so either you have the new or the old.
 
> So I imagine this pull request will be closed when we close #1017 , correct?

Correct, it is included in the combined patchset.

jeremiah-c-leary (approved):

alonbl:
Closed via #1017

PR #1016, opened 2023-10-08
Support file configuration without scanning the file

alonbl:
**Description**

Currently per file configuration is specified under `file_list`, the `file_list` has side effect of scanning the file regardless of what files provided in command-line.

Configuration should be able to specify per file configuration without adding the file to the scan.

New implementation will look for the file under `file_rules` and merge it to the existing `file_list` configuration to keep backward compatibility.

### DIFF
--- a/vsg/tests/vsg/test_vsg.py
+++ b/vsg/tests/vsg/test_vsg.py
@@ -520,5 +520,3 @@ class testVsg(unittest.TestCase):
             iExitStatus = e.returncode
 
         self.assertEqual(lActual, lExpected)
-
-
--- a/vsg/tests/vsg/test_vsg.py
+++ b/vsg/tests/vsg/test_vsg.py
@@ -481,3 +481,44 @@ class testVsg(unittest.TestCase):
 
         self.assertEqual(utils.replace_total_count_summary(lActualStdErr), lExpectedStdErr)
         self.assertEqual(utils.replace_total_count_summary(lActualStdOut), lExpectedStdOut)
+
+    def test_globbing_filenames_in_configuration_with_file_rules(self):
+        lExpected = []
+        lExpected.append('ERROR: vsg/tests/vsg/entity2.vhd(8)port_008 -- Change number of spaces after *out* to 3.')
+        lExpected.append('')
+
+        try:
+            subprocess.check_output(['bin/vsg','--configuration','vsg/tests/vsg/config_glob_with_file_rules.yaml','--output_format','syntastic'])
+        except subprocess.CalledProcessError as e:
+            lActual = str(e.output.decode('utf-8')).split('\n')
+            iExitStatus = e.returncode
+
+        self.assertEqual(lActual, lExpected)
+
+    def test_configuration_with_file_rules_and_no_file_list_entity2(self):
+        lExpected = []
+        lExpected.append('ERROR: vsg/tests/vsg/entity2.vhd(8)port_008 -- Change number of spaces after *out* to 3.')
+        lExpected.append('')
+
+        try:
+            subprocess.check_output(['bin/vsg','--configuration','vsg/tests/vsg/config_file_rules.yaml', '-f', 'vsg/tests/vsg/entity2.vhd', '--output_format','syntastic'])
+        except subprocess.CalledProcessError as e:
+            lActual = str(e.output.decode('utf-8')).split('\n')
+            iExitStatus = e.returncode
+
+        self.assertEqual(lActual, lExpected)
+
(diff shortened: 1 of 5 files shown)


alonbl:
I am unsure why adding a simple loop is lowering the score and fails the codeclimate, if you want I can split this into two functions.

My patches are rebased and ready without conflicts at https://github.com/alonbl/vhdl-style-guide/tree/alonbl

jeremiah-c-leary:
Good Morning @alonbl ,

I just wanted to ensure I understand your use case.  From your description, it sounds like splitting the files to analyze using `file_list` and then setting specific file configurations via `file_rules` would allow you to define file specific rule exceptions without forcing that file to be analyzed.

I imagine you could define all the rule exceptions using `file_rules` and then if you analyze a file via the command line those exceptions would be applied.  So something like this:
```yaml
#config.yaml
file_rules:
  - file_a.vhd:
       rule:
          constant_001:
            disable: True
```
If I passed `file_a.vhd` to `vsg` like this:
```text
vsg -c config.yaml -f file_a.vhd
```
Then the `file_rules` would be applied to `file_a.vhd`
However, if I just ran `vsg` with just the configuration:
```text
vsg -c config.yaml
```
No files would be analyzed because `file_list` does not exist.

Does that cover you use case?

At the moment it is not clear what advantage it provides, but it could just me being dense this morning.  Could you elaborate on how you intend to use `file_list` and `file_rules`?

Regards,

--Jeremy

alonbl:
Hello @jeremiah-c-leary,

Yes, this exactly the use case, I left the rules within `file_list` only for backward compatibility, it may be removed in future versions, leaving only `file_rules` to specify exceptions.

Currently, the `file_list` exceptions inject the file into the scan which is undesired in many cases.

Thanks,

alonbl:
Rebased after last merge, you can approve this series and then look into https://github.com/jeremiah-c-leary/vhdl-style-guide/pull/1017 as I resolved all conflicts there.

jeremiah-c-leary:
I like this idea.  I think it might solve a situation I ran into earlier this year.

I had a configuration that looked something like this:
```yaml
#config.yaml
file_list:
   - ../**/*.vhd
```
since I did not want to maintain a yet another list of files I used globbing.  However, I could not effectively apply rule exceptions.  Although I suppose I could have just listed the files after the glob.  Anyway, having a separate `file_rules` section to the configuration would allow for applying rules on a per file basis while also globbing:
```yaml
#config.yaml
file_list:
   - ../**/*.vhd
file_rules:
   - ../interfaces/usb.vhd
        rule:
           constant_001:
                disable: True
```
Let me see if I can add some tests for `file_rules` before we merge it to master.

Regards,

--Jeremy

jeremiah-c-leary:
Hey @alonbl ,

I added three tests and updated the documentation for `file_rules` and then pushed it to your branch.  Could you double check my changes?

Thanks,

--Jeremy

vsg/tests/vsg/test_vsg.py
```diff
@@ -481,3 +481,44 @@ def test_summary_output_format_multiple_mixed_jobs_2(self):
...
+
+        try:
+            subprocess.check_output(['bin/vsg','--configuration','vsg/tests/vsg/config_file_rules.yaml', '-f', 'vsg/tests/vsg/entity1.vhd', '--output_format','syntastic'])
+        except subprocess.CalledProcessError as e:
+            lActual = str(e.output.decode('utf-8')).split('\n')
+            iExitStatus = e.returncode
+
+        self.assertEqual(lActual, lExpected)
+
+
```
alonbl, line 524: @jeremiah-c-leary works nicely!

these trailing lines are on purpose?

jeremiah-c-leary: @alonbl ,

Those trailing lines are not on purpose.  I have committed an update to remove them.

I also refactored `configure_rules` to reduce the complexity, and hopefully make it easier to read.

--Jeremy

alonbl: Thank you!

The refactor is subjective :)
For me, splitting the code into so many functions for a single task is less easier to read, but that's not important.
Please review my other patches ;)

jeremiah-c-leary:
@alonbl ,

I will plan to merge this to master early this week unless you find something.

Thanks again for the PR.  These quality of life updates will really help users.

--Jeremy

alonbl:
> unless you find something.

Tested and is working find in my environment.

jeremiah-c-leary:
Awesome, I will merge it now.

--Jeremy

PR #1055, opened 2023-11-30
Issue#1053:  Adding rule to check for reserved words.

### LINKED ISSUE #1053: Report VHDL Reserved Words
The VHDL LRM has a short list (I count 120) of reserved keywords that identifiers (signals, functions, labels, etc) cannot have - "signal", "port", "bus", "default", etc. Each new revision generally adds a few new words - "protected", "private", "context" etc. (I also count 4 words removed in the history of the standard)

At present nothing in vsg nor vscode alerts me that a given identifier has been given a reserved name, which is complicating my use-case - "make this VHDL-1993-ish code compile cleanly under VHDL-2008". For now I have a workaround of trying to compile the code in ModelSim, which is pretty good about protesting about the new keywords _that it knows about_, but a linter message would be nice too.

A useful feature would be a rule that checks each identifier is not on the set of reserved words. For myself, any word reserved from 1987 to 2019 is sufficient; but other users might appreciate "only test words from VHDL-2002" if that is the standard their project is being compiled to.

It is also worth noting that functions can sometimes be called operations - such as "nand", "xor", when overloading the defined operations; but that naming a signal "xor" is more likely an error.

Thanks,

jeremiah-c-leary:
Resolves #1053


SittingDuc:
rules.py line 26, set `dMap['2008']` has some repeated symbols - 'assert', 'vunit', and a few others. Probably a copy-pasta while reading the pdf

PR #1129, opened 2024-02-19
Fix typo and quote relativeFile

MDXZ-delti:
"autodetect" is not accepted, while "autoDetect" is.
Without quotes, an error is thrown if relativeFile contains spaces.


jeremiah-c-leary:
Morning @MDXZ-delti ,

Thank you for the update.  I will get it merged to master.

Regards,

--Jeremy

PR #1154, opened 2024-03-25
docs: update installation methods

alonbl:
build system was modified to tox, documentation should have been updated.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -144,11 +144,17 @@ The latest development version can be cloned...
 
     git clone https://github.com/jeremiah-c-leary/vhdl-style-guide.git
 
-...and then installed locally...
+Install prerequisites.
 
 .. code-block:: bash
 
-    python setup.py install
+    pip install tox
+
+Build locally, artifacts will appear in ``dist`` directory.
+
+.. code-block:: bash
+
+    tox
 
 Usage
 -----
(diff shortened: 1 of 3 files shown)


alonbl:
@jeremiah-c-leary there is a strange underline in GitHub rendering, from Installation and on, I could not find why, even before the place I touched.

jeremiah-c-leary:
Evening @alonbl ,

Is there a reason a user would install `tox` if they did not want to do any development?  There is a section of the documentation where installing `tox` could be added.

You can see the gyrations I went through in my release process.  You're method is a lot cleaner.  Thank you.

--Jeremy

alonbl:
> Is there a reason a user would install `tox` if they did not want to do any development? There is a section of the documentation where installing `tox` could be added.

As you provide the `bin` directory, users may checkout and use the tool directly as you intended.
But why will they want to use the master if they are not developing?
My intention was to remove all leftovers of the old build system which I have missed, feel free to modify in any way you like, I do not understand what section of documentation you refer to.

jeremiah-c-leary:
Evening @alonbl ,

> But why will they want to use the master if they are not developing?

That is a good point.

> My intention was to remove all leftovers of the old build system which I have missed, feel free to modify in any way you like, I do not understand what section of documentation you refer to.

I already have a issue to update the development section, so I will make any necessary changes there.

I will merge this to master.

--Jeremy

PR #1159, opened 2024-03-29
Issue#1156:  Changed prefix/suffix exceptions to allow for regular expressions.

### LINKED ISSUE #1156: Subtype exceptions
Hi,

I'm working in a project which has enabled the prefix rules for (sub)types. All (sub)types must be prefixed with `t_`. That works well but there is an exception. When specifying a field in a register we use the following construct:

subtype A_FIELD is integer range 7 downto 0;

While `A_FIELD` is a subtype, it is not used as such and we would like to have an exception. From what I can tell, this the prefix rule for subtypes doesn't support exceptions and even if it did, does exceptions support regexps? Something like `exception = ['*FIELD']?

As a workaround I can:

1. use `vsg_on` and `vsg_off`
2. prefix the subtype with `t_` and then have an alias for the subtype without the prefix.

What I would like is a more direct way to deal with this in the VSG configuration

jeremiah-c-leary:
Resolves #1156

### DIFF
--- a/docs/configuring_prefix_and_suffix_rules.rst
+++ b/docs/configuring_prefix_and_suffix_rules.rst
@@ -54,10 +54,10 @@ Exceptions to enforcing prefixes and suffixes can be given using the :code:`exce
             # Each suffix rule needs to be enabled explicitly.
             disable: false
             suffixes: ['_i', '_o']
-            exceptions: ['CLK']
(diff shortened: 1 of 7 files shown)


LarsAsplund:
Thanks @jeremiah-c-leary, this works perfectly!

PR #1183, opened 2024-06-19
Fix bug in whitespace between tokens and add test for it

edgarmv:
**Description**
Fixes two bugs in `whitespace_between_tokens.py`. The first bug is passing the number of spaces as index into `lTokens` instead of the `num` parameter. This fails when `number_of_spaces` is more than 1,  and is silently catched by an `except IndexError`.

The second bug is fixed by adding extraction of `number_of_spaces` if the number ends with a  `+`.  There already is a function for analyzing if it exists, but the functionality of extracting it is missing. 

Also add a test in `element_association` that checks that you can set `number_of_spaces: '2+'`, covering both these bugs.

### DIFF
--- a/tests/element_association/rule_100_test_input.fixed_spaces_gte2.vhd
+++ b/tests/element_association/rule_100_test_input.fixed_spaces_gte2.vhd
@@ -1,0 +1,29 @@
+
+architecture rtl of fifo is
+
+begin
+
+  a <= (others  => (others  => '0'));
+
+  process begin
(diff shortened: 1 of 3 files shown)


jeremiah-c-leary:
Evening @edgarmv ,

Thank you very much for finding these issues.  I also appreciate that you added a test to catch the error along with the bug fix.

I will merge this to master.

Regards,

--Jeremy

PR #1208, opened 2024-07-27
Issue#1207:  Fixing default values in multiline indent documentation.

### LINKED ISSUE #1207: Documentation states incorrect option defaults for multiline indent rules
**Environment**
```
$ vsg --version
VHDL Style Guide (VSG) version: 3.25.0
```
Installed with Pip on Ubuntu 24.04.

**Describe the bug**
[The documentation](https://github.com/jeremiah-c-leary/vhdl-style-guide/blob/3.25.0/docs/configuring_multiline_indent_rules.rst) states that the multiline indent rules have the following defaults: `yes` for option `align_left` and `no` for option `align_paren`. (These are the defaults that I prefer.) However, this documentation does not appear to be accurate as in practice the rule defaults appear to be the other way around: `no` for option `align_left` and `yes` for option `align_paren`.

**To Reproduce**
1. Run `vsg -oc config`
2. Open file `config`
3. Note that the rules that have `align_left` and `align_paren` for options have values different from the documented defaults.

**Expected behavior**
The default options to match the documentation.

jeremiah-c-leary:
Resolves #1207

### DIFF
--- a/docs/configuring_multiline_indent_rules.rst
+++ b/docs/configuring_multiline_indent_rules.rst
@@ -38,10 +38,10 @@ There are several options to these rules:
 +---------------+----------+---------------+----------------------+
 | Option        | Values   | Default       | Description          |
 +===============+==========+===============+======================+
-| |align_left|  | |values| | |default_yes| | * |align_left__yes|  |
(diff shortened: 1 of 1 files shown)


JvWesterveld:
LGTM!

PR #1210, opened 2024-08-12
vsg: print file name when ClassifyError

alonbl:
In current implementation the user see something like the following when `ClassifyError` is raised:

```
    Error: Unexpected token detected while parsing procedure_call_statement @ Line 49, Column 22 in file None
           Expecting : ;
           Found     : )
```

In a project that contains multiple files, the user is clueless as he does not know which of the files caused the error.

This change adds the name of the file to the message.

```
    Error while processing xxx:
    Error: Unexpected token detected while parsing procedure_call_statement @ Line 49, Column 22 in file None
           Expecting : ;
           Found     : )
````

### DIFF
--- a/tests/vsg/test_main.py
+++ b/tests/vsg/test_main.py
@@ -167,13 +167,14 @@ class testMain(unittest.TestCase):
     @mock.patch("sys.stderr")
     def test_junit_with_file_that_fails_to_parse(self, mock_stderr):
         lStdErr = []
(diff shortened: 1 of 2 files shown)


jeremiah-c-leary:
Evening @alonbl ,

Thanks for the update with the file name.  That was something I always wanted to do but never seemed to find the time.

I will get this merged to master for the next release.

Regards,

--Jeremy

PR #1217, opened 2024-08-25
Issue#1216: Added escape characters for "**" in docs.

### LINKED ISSUE #1216: The ** in the description for whitespace_011 render as a URL and generates a warning when building docs.
**Environment**
VSG v3.25.0

**Describe the bug**
image
The description for whitespace_011 includes the string "**". The restructuredtext parser recognises this as the start of a **strong text** section, and does not find an end to that strong text section. This causes it to render as a hyperlink on the docs for some reason.

**To Reproduce**
1. Go to https://vhdl-style-guide.readthedocs.io/en/stable/whitespace_rules.html#whitespace-011
2. See error
3. Build the docs
4. See warning
```
<path>/docs/whitespace_rules.rst:201: WARNING: Inline strong start-string without end-string. [docutils]
```

**Expected behavior**
This should be rendered as plain text

JHertz5:
Resolves #1216.

### DIFF
--- a/docs/whitespace_rules.rst
+++ b/docs/whitespace_rules.rst
@@ -198,7 +198,7 @@ whitespace_011
 
 |phase_2| |error| |whitespace|
 
-This rule checks for at least a single space before and after math operators +, -, /, * and **.
+This rule checks for at least a single space before and after math operators +, -, /, * and \*\*.
 
 **Violation**
 
--- a/vsg/rules/whitespace/rule_011.py
+++ b/vsg/rules/whitespace/rule_011.py
@@ -13,7 +13,7 @@ lTokens.append(token.miscellaneous_operator.double_star)
 
 class rule_011(n_spaces_before_and_after_tokens):
     """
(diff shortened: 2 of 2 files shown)


JHertz5:
It might be of a stretch for every reader to see `/*/*` in the python comment and know that it means `**`. An alternative solution that I have in mind it to use backticks "``" to put the operators in code formatting, but this is inconsistent with the rest of the documentation. We could also just remove the list of mathematical operators from the comment, as per [whitespace_013](https://vhdl-style-guide.readthedocs.io/en/latest/whitespace_rules.html#whitespace-013 )and [whitespace_100](https://vhdl-style-guide.readthedocs.io/en/latest/whitespace_rules.html#whitespace-100), which do not list the logical and relational operators.

jeremiah-c-leary:
I think I may end up dropping code climate.  Due to the similarities in the code I get a lot of duplicate code warnings so I end up ignoring the code climate report.  I suppose I could turn those off though....

PR #1226, opened 2024-09-07
Issue-1225 vsg_rule_doc_gen script does not run.

### LINKED ISSUE #1225: vsg_rule_doc_gen script does not run
**Environment**
v3.25.0

**Describe the bug**
The vsg_rule_doc_gen script runs the function `create_rule_documentation`, which has a single required argument `path`. Unfortunately, this script does not accept any arguments (other than `-h`) so it is not possible to provide a path, and the code within the `main` function for this script does not pass any argument to `create_rule_documentation`.

**To Reproduce**
1. Run `./bin/vsg_rule_doc_gen `
2. See error
```
$ ./bin/vsg_rule_doc_gen 
Traceback (most recent call last):
  File "/home/jhertzog/code/vhdl-style-guide-fork/./bin/vsg_rule_doc_gen", line 16, in <module>
    main()
  File "/home/jhertzog/code/vhdl-style-guide-fork/bin/../vsg/__rule_doc_gen__.py", line 24, in main
    create_rule_documentation()
TypeError: create_rule_documentation() missing 1 required positional argument: 'path'
```

**Expected behavior**
I expect the script to run and to update the documentation files.

JHertz5:
Resolves #1225.

I've added a very rudimentary test for the vsg_rule_doc_gen script. It basically just checks that:
- The script runs without crashing.
- The script produces output files to the user-specified directory.
- The script doesn't print anything to stdout.

I haven't done any further in-depth testing since the functionality is very thoroughly tested in test_rule_doc. If you'd like the testing to be more thorough, I'd be happy to work on it a bit more, but, considering that the top-level script wasn't tested at all before, hopefully it's good enough.

### DIFF
--- a/docs/ieee_rules.rst
+++ b/docs/ieee_rules.rst
@@ -1,4 +1,3 @@
-
 .. include:: includes.rst
 
 IEEE Rules
--- a/docs/reserved_rules.rst
+++ b/docs/reserved_rules.rst
@@ -6,7 +6,7 @@ Reserved Rules
 reserved_001
 ############
 
-|phase_1| |error| |structure|
+|phase_1| |error| |unfixable| |structure|
 
 This rule checks for VHDL reserved words being used as identifiers and names.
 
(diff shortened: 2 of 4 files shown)


jeremiah-c-leary:
@JHertz5 ,

I think at this point just having the script run is good enough.

If there are requests in the future to improve it, I can update it then.

Thanks again,

--Jeremy

PR #1233, opened 2024-09-28
Issue#1232: Replaced ieee_500 with type_mark_500

### LINKED ISSUE #1232: Capitalisation rules for type identifiers do not work on types that are declared in a different file.
**Environment**
VSG v3.26.0

**Describe the bug**
The rule `type_004` enforces capitalisation of type identifiers. This rule appears to be triggered by the type declaration. As a result, this means that types declared elsewhere, e.g. in a package that is a separate file, or that is part of the standard library, such as `character` are not corrected by this rule.

I suspect that the latter example is the reason that the rule `ieee_500` was introduced. This rule solves the problem for a few predefined types, namely:
- `std_logic_vector`
- `std_ulogic_vector`
- `std_ulogic`
- `std_logic`
- `integer`
- `signed`
- `unsigned`

but it leaves out many more, including `natural` which is defined by the parser, but isn't included in the list of tokens that `ieee_500` acts upon. There are also many other types like `string`,  `character`, etc. that aren't included in the terms recognised by the parser. My point is that if we could try to add every predefined type, but that just adds more code and puts us at a greater risk of introducing human errors.

Instead, I suggest that, unless there is another reason that these keywords are treated as special, we treat them as any other type name and handle the capitalisation of type and subtype names in a more generic way..

One thing that we could do is replace rule `ieee_005` with a single rule for `type_mark` tokens. That way, `type_004` and `subtype_002` will continue to handle types and subtype with local declarations, and the new rule, `type_mark_500`, (or perhaps a better name), could handle type marks for declared elsewhere, including VHDL's predefined types.

**To Reproduce**
1. Create a file "test.vhd" with the following contents:
```vhdl
library ieee;
  use ieee.std_logic_1164.all;
  use ieee.numeric_std.all;
  use ieee.math_real.all;

entity test is
end entity test;

architecture rtl of test is

  type my_type is range 10 to 100;

  subtype my_subtype   is MY_TYPE range 10 to 20;
  subtype my_subtype_t is MY_EXTERNAL_TYPE range 10 to 20;
  subtype my_int       is INTEGER range 10 to 100;
  subtype my_nat       is NATURAL range 10 to 100;
  subtype my_pos       is POSITIVE range 10 to 100;
  subtype my_char      is CHARACTER range 'a' downto 'b';

  signal s_my_subtype   : MY_EXTERNAL_SUBTYPE;
  signal s_my_subtype   : MY_SUBTYPE;
  signal s_my_subtype_t : MY_SUBTYPE_T;
  signal s_my_int       : MY_INT;
  signal s_my_nat       : MY_NAT;
  signal s_my_pos       : MY_POS;
  signal s_my_char      : MY_CHAR;
  signal s_my_type_t    : MY_TYPE_T;
  signal s_int          : INTEGER;
  signal s_nat          : NATURAL;
  signal s_pos          : POSITIVE;
  signal s_char         : CHARACTER;

begin

end architecture rtl;
```
2. Run `vsg -f test.vhd --fix`
3. Observe test.vhd:
```vhdl
library ieee;
  use ieee.std_logic_1164.all;
  use ieee.numeric_std.all;
  use ieee.math_real.all;

entity test is
end entity test;

architecture rtl of test is

  type my_type is range 10 to 100;

  subtype my_subtype   is my_type range 10 to 20;
  subtype my_subtype_t is MY_EXTERNAL_TYPE range 10 to 20;
  subtype my_int       is integer range 10 to 100;
  subtype my_nat       is NATURAL range 10 to 100;
  subtype my_pos       is POSITIVE range 10 to 100;
  subtype my_char      is CHARACTER range 'a' downto 'b';

  signal s_my_subtype   : MY_EXTERNAL_SUBTYPE;
  signal s_my_subtype   : my_subtype;
  signal s_my_subtype_t : my_subtype_t;
  signal s_my_int       : my_int;
  signal s_my_nat       : my_nat;
  signal s_my_pos       : my_pos;
  signal s_my_char      : my_char;
  signal s_my_type_t    : MY_TYPE_T;
  signal s_int          : integer;
  signal s_nat          : NATURAL;
  signal s_pos          : POSITIVE;
  signal s_char         : CHARACTER;

begin

end architecture rtl;
```
4. Note that `MY_EXTERNAL_TYPE `, `MY_EXTERNAL_SUBTYPE `, `NATURAL`, `POSITIVE`, and `CHARACTER` are still upper case. The latter three are all predefined VHDL types.

**Expected behavior**
I expect the tool to enforce capitalisation on all type marks consistently.

JHertz5:
Resolves #1232.

Potential issues:
- ~~Conflict/overlap with `type_014` and `subtype_002`. Those rules enforce _consistent_ capitalisation, rather than strictly upper or lower case, so we would lose the ability to allow for more complex case conventions if we disable them. However, from the new rule, we gain the ability to enforce casing on more type marks. For the purposes of this PR, I have assumed that we will get around this problem by setting `type_014` and `subtype_002` to be disabled by default, but I am very open to feedback on this.~~ This has been resolved. The new rule only enforces capitalisation for type marks which are not declared in the file.
- I am have removed the parsing of ieee/standard types as special tokens to avoid overlap/conflicts with `ieee_500`, and to let the new rule cover those instances. Happy for this to be reverted if this is not preferred.
- I am also open to feedback on the name "type_mark". It comes from the LRM, but I suspect that the majority of users would not be familiar with this term. Perhaps "type_name" would be more commonly known?

Basically, any feedback/suggestions/advice is/are welcome, and I'm happy for this PR to be rejected/heavily edited if we don't like this change.


tests/rule_group/test_input.config_1.fixed.vhd
```diff
@@ -55,7 +55,7 @@ PACKAGE some_pkg IS
 
   PROCEDURE proc_1;
 
-  FUNCTION func_1 RETURN INTEGER;
+  FUNCTION func_1 RETURN Integer;
```
jeremiah-c-leary, line 58: Seems like `Integer` should be lowercase because it is a `type_mark`:
```mono
function_specification ::=
   [ pure | impure ] function designator
    subprogram_header
    [ [ parameter ] ( formal_parameter_list ) ] return type_mark
```

tests/rule_group/test_input.config_1.actual.vhd
jeremiah-c-leary, line 1: This file can be deleted.  It is left over from running the test and should not have been committed.

tests/rule_group/test_input.config_2.actual.vhd
jeremiah-c-leary, line 1: This file can be deleted.  It is left over from running the test and should not have been committed.

tests/rule_group/test_input.config_3.actual.vhd
jeremiah-c-leary, line 1: This file can be deleted.  It is left over from running the test and should not have been committed.

tests/rule_group/test_input.config_2.fixed.vhd
```diff
@@ -55,7 +55,7 @@ package SOME_PKG is
 
   procedure PROC_1;
 
-  function FUNC_1 return integer;
+  function FUNC_1 return Integer;
```
jeremiah-c-leary, line 58: Similar to the other test, I believe `Integer` should be uppercase.

tests/rule_group/test_input.config_3.fixed.vhd
```diff
@@ -55,7 +55,7 @@ package SOME_PKG is
 
   procedure PROC_1;
 
-  function FUNC_1 return integer;
+  function FUNC_1 return Integer;
```
jeremiah-c-leary, line 58: Similar to the other tests, I believe `Interger` should be uppercase.

jeremiah-c-leary (changes requested):
After looking through all the classification files, I think this PR was a good idea.  The IEEE_500 rule wasn't really thought out too well, so the fact it is deprecated is justified.

I appreciate all the work on this PR.  83 files is a significant effort.

I just had a couple of questions on some files.

Thanks again,

--Jeremy

tests/rule_group/test_input.config_1.actual.vhd
JHertz5: That makes sense, I was a bit confused about why there were two identical files (`fixed` and `actual`) but didn't think about it very hard! I'll remove all of those `actual` files.

tests/rule_group/test_input.config_2.actual.vhd
JHertz5: That makes sense, I've removed the file.

tests/rule_group/test_input.config_3.actual.vhd
JHertz5: That makes sense, I've removed the file.

tests/rule_group/test_input.config_1.fixed.vhd
JHertz5: Hi @jeremiah-c-leary. You are absolutely correct. You've caught a flaw in the PR, `type_mark_500` does not work in packages or package bodies.

This is a consequence of my use of the consistent case rules to inform whether a `type_mark` has a local definition. This is because the consistent case rules also do not work in packages or package bodies. I can provide the following test as evidence:

Source file, test.vhd:
```vhdl

package test is

  type My_type is (a1, b2, c3);

  function my_func1 return MY_TYPE; -- Not fixed by consistent case rule

  function my_func2 return MY_EXTERNAL_TYPE; -- Not fixed by type_mark_500

end package test;

package body test is

  function my_func1 return MY_TYPE is -- Not fixed by consistent case rule
  begin
    return A1;
  end function my_func1;

  function my_func2 return MY_EXTERNAL_TYPE is -- Not fixed by type_mark_500
  begin
    return A1;
  end function my_func2;

end package body test;

architecture rtl of test is

  subtype My_subtype is my_type range a1 to b2; -- Not fixed by consistent case rule, but my_type is made lowercase by type_mark_500

  function my_func1 return my_type is -- Not fixed by consistent case rule, but made lowercase by type_mark_500
  begin
    return A1;
  end function my_func1;

  function my_func2 return my_external_type is -- Not fixed by type_mark_500, but made lowercase by type_mark_500
  begin
    return A1;
  end function my_func2;

  function my_func3 return My_subtype is -- Fixed by consistent case rule (because definition is within architecture)
  begin
    return A1;
  end function my_func3;

begin

end architecture rtl;
```
Config file, test.yml:
```yaml
rule:
  # To allow types and subtypes to be non-lowercase.
  type_004:
    disable: yes
  subtype_501:
    disable: yes
  # To make the example code more compact for readability.
  subprogram_body_203:
    disable: yes
  subprogram_body_204:
    disable: yes
```
Output (running on this PR's branch):
```
$ bin/vsg -f test.vhd -c test.yml -ap
File:  test.vhd
Phase 7 of 7... Reporting
Total Rules Checked: 851
Total Violations:    0
  Error   :     0
  Warning :     0
```

From what I can see, the solution to the problem is to correct the consistent case rules to work in packages and package bodies, which will in turn correct `type_mark_500`. I could do that work as part of this PR, but if we want to avoid growing the PR further, I could raise a separate issue instead. I would prefer the latter route, but I'm happy to go either way. Do you have a preference?

Thanks,
Jukka

tests/rule_group/test_input.config_2.fixed.vhd
JHertz5: I've left a response in this comment https://github.com/jeremiah-c-leary/vhdl-style-guide/pull/1233#discussion_r1900409679

tests/rule_group/test_input.config_3.fixed.vhd
JHertz5: I've left a response in this comment https://github.com/jeremiah-c-leary/vhdl-style-guide/pull/1233#discussion_r1900409679.

tests/rule_group/test_input.config_1.fixed.vhd
jeremiah-c-leary: Morning @JHertz5 ,

> From what I can see, the solution to the problem is to correct the consistent case rules to work in packages and package bodies, which will in turn correct type_mark_500. I could do that work as part of this PR, but if we want to avoid growing the PR further, I could raise a separate issue instead. I would prefer the latter route, but I'm happy to go either way. Do you have a preference?

I think a separate issue would be preferred.  It would be more focused on consistent case rules in packages versus this issue where we are replacing ieee_500 with type_mark_500.

I will review the current PR.

Thanks,

--Jeremy

jeremiah-c-leary (approved):
PR looks good.  We can address the remaining comments in a different issue.

tests/rule_group/test_input.config_3.fixed.vhd
jeremiah-c-leary: See response on #1233

tests/rule_group/test_input.config_2.fixed.vhd
jeremiah-c-leary: See response in #1233

tests/rule_group/test_input.config_1.actual.vhd
jeremiah-c-leary: I was confused about those files also.  The `actual` in the file name jogged my memory.

tests/rule_group/test_input.config_1.fixed.vhd
JHertz5: Hi @jeremiah-c-leary

Excellent, thanks very much! I've raised #1384 to capture the consistent case changes.

Jukka

PR #1237, opened 2024-10-05
Issue#1236: Rule for case in hexadecimal constants and base specifier

### LINKED ISSUE #1236: Rule for case in hexadecimal constants and base specifier
Our coding guideline prescribes upper case for letters in hexadecimal numeric constants and lower case for the "x" before the hex string. I cannot find a rule that allows enforcing this:

Before:
`constant c : std_logic_vector(31 downto 0) := X"cafebabe";`

Required after:
`constant c : std_logic_vector(31 downto 0) := x"CAFEBABE";`

Add a rule that allows enforcing the case of letters in hexadecimal numeric constants (`extended_digit`) and, separately, the `base_specifier`.

JHertz5:
Resolves #1236.

Potential issues to consider for the reviewer:
- The new rules are part of the case group, but they are not part of any case subgroup. From what I can tell, all other case rules are part of a subgroup, but it didn't seem like these rules fit into any of those groups.
- I've set the default value of rule_501 to "upper". I'm not sure whether there are any other case rule where the default is "upper",  if this is not preferred, we can change the default to match the other case rules.
- It's my first attempt at modifying the classification/tokenisation rules, so I may have done things in a way that is suboptimal. There is probably some room for improvement there, so I'd welcome any suggestions.
- From what I can tell, this is the first time that the classifiers are classifying items that don't match a fixed string (e.g. "constant"). I've introduced regular expressions into vhdlFile/util.py, with functions like `matches_next_token` as an alternative to `is_next_token` that matches a pattern instead of a fixed string. Maybe there is a better/cleaner way to implement this.

### DIFF
--- a/docs/bit_string_literal_rules.rst
+++ b/docs/bit_string_literal_rules.rst
@@ -34,6 +34,8 @@ This rule checks the bit value has proper case.
 
 |configuring_uppercase_and_lowercase_rules_link|
 
+The default style is :code:`upper`.
+
 **Violation**
 
 .. code-block:: vhdl
--- a/docs/bit_string_literal_rules.rst
+++ b/docs/bit_string_literal_rules.rst
@@ -1,0 +1,48 @@
+.. include:: includes.rst
+
+Bit String Literal Rules
+------------------------
+
+bit_string_literal_500
+######################
+
+|phase_6| |error| |case|
+
+This rule checks the base specifier has proper case.
+
+|configuring_uppercase_and_lowercase_rules_link|
+
+**Violation**
+
+.. code-block:: vhdl
+
+    signal test : my_vector := X"FFF";
+
+
+**Fix**
+
+.. code-block:: vhdl
+
+   signal test : my_vector := x"FFF";
+
+bit_string_literal_501
+######################
+
+|phase_6| |error| |case|
+
+This rule checks the bit value has proper case.
+
+|configuring_uppercase_and_lowercase_rules_link|
+
+**Violation**
+
+.. code-block:: vhdl
+
+    signal test : my_vector := x"FFF";
+
+
+**Fix**
+
+.. code-block:: vhdl
+
+   signal test : my_vector := x"fff";
--- a/docs/configuring_uppercase_and_lowercase_rules.rst
+++ b/docs/configuring_uppercase_and_lowercase_rules.rst
@@ -251,6 +251,9 @@ Rules Enforcing Case
 * `attribute_specification_502 <attribute_specification_rules.html#attribute-specification-502>`_
 * `attribute_specification_503 <attribute_specification_rules.html#attribute-specification-503>`_
 
+* `bit_string_literal_500 <bit_string_literal_rules.html#bit-string-literal-500>`_
+* `bit_string_literal_501 <bit_string_literal_rules.html#bit-string-literal-501>`_
+
 * `block_500 <block_rules.html#block-500>`_
 * `block_501 <block_rules.html#block-501>`_
 * `block_502 <block_rules.html#block-502>`_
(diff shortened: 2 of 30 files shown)


jeremiah-c-leary:
Afternoon @JHertz5,

You are becoming quite the expert.   With this one pull request you have touched the entire "stack", from the lexer, parser and all the way through a new rule.  I am very impressed.  You even added the new rule class to the `vsg/rules/__init__.py` file, which I manage to miss occasionally.

I did not see this file updated though:  `docs/rules.rst` which will bring the new rule document into the document build process.

> The new rules are part of the case, but they are not part of any case subgroup. From what I can tell, all other case rules are part of a subgroup, but it didn't seem like these rules fit into any of those groups.

I would agree with your assessment.  It is not a label, keyword or a name.   I do not think it is worth it to create another subgroup either.  That could always be done at a later date if needed.

> I've set the default value of rule_501 to "upper". I'm not sure whether there are any other case rule where the default is "upper", if this is not preferred, we can change the default to match the other case rules.

All the other case rules default to lower, although it is hard for me to see setting `rule_501` to `lower`.  So I think keeping it at `upper` makes sense.  Hex looks really odd to me when it is lower case.

> It's my first attempt at modifying the classification/tokenisation rules, so I may have done things in a way that is suboptimal. There is probably some room for improvement there, so I'd welcome any suggestions.

I think your code updates look great.  I like how you kept the same naming conventions and coding styles.  It looks cohesive with the rest of the code.  Except for this bit in `expression.py`:
```python
            for oToken in [external_name, bit_string_literal]:
                iCurrent = oToken.detect(iCurrent, lObjects)
                if iCurrent != iPrevious:
                    continue
```
Which I never would have thought to code in that manner.  I would have used an `if elsif` structure to check each option.  I could have used that in some classification files where multiple options exist.  I will have to keep this in mind if I ever need to touch those files.  So thank you for the code improvement idea.

> From what I can tell, this is the first time that the classifiers are classifying items that don't match a fixed string (e.g. "constant"). I've introduced regular expressions into vhdlFile/util.py, with functions like matches_next_token as an alternative to is_next_token that matches a pattern instead of a fixed string. Maybe there is a better/cleaner way to implement this.

I would agree this is the first instance of using a non constant check.  The only improvement I can think of would be to change the strings in `vsg/vhdlFile/classify/bit_string_literal.py`:
```python
sIntegerPattern = r"^\d+$"
sBaseSpecifierPattern = r"^(([us]?[box])|d)$"
sBitValueStringPattern = r"^\"[0-9a-fhluwxz\-_]*\"$"
```
into compiled regular expression objects and then pass the objects into the checking function.  Check out the file `vsg/rules/case_utils.py` where regular expressions are used to check for `camelCase` and `PascalCase`.  I believe if you do not use `re.compile`, then the regular expression will need to be compiled each time it is called.  It should hopefully be a small change to the new functions you added.

Overall this is a fantastic pull request.  You have seen probably 90% of the code base and have a good understand of how everything fits together.  I really appreciate your effort.

Regards,

--Jeremy

JHertz5:
Hi @jeremiah-c-leary.
Thanks for your kind words! I've really enjoyed exploring your excellent codebase so far and learning more about python and VHDL in the process, so I'd like to thank you as well. I will take your feedback onboard and update the PR soon.
Thanks again!
Jukka

JHertz5:
Hi @jeremiah-c-leary,

Thanks again for your feedback!

> I did not see this file updated though: docs/rules.rst which will bring the new rule document into the document build process.

You're right, I completely forgot to list the new rule group! Thanks for catching that.

>The only improvement I can think of would be to change the strings in vsg/vhdlFile/classify/bit_string_literal.py:
into compiled regular expression objects and then pass the objects into the checking function. 

That is interesting! I've not use regex and python together before, so compiled regular expressions are new to me. It definitely makes sense as an efficiency improvement, and it's made the code slightly simpler too. Thanks very much for that suggestion!

I believe that I have actioned all of your feedback, but if there's anything else that you'd like tweaked or modified, please let me know.

Regards,
Jukka

jeremiah-c-leary:
Evening @JHertz5 ,

Everything looks good.  I will merge this to master.

Again, thank you for the contribution.

Regards,

--Jeremy

PR #1240, opened 2024-10-06
Add rules to require **begin** keyword for generate statements

larshb:
As discussed in #1239 I think these would be a nice addition to the rules.

I wanted to add an accompanying `case_generate_alternative_001`, but this was a bit more tricky to figure out.

I guess we'd need a new base-rule e.g. `insert_token_right_of_token_if_it_does_not_exist_before_possible_tokens` to list the following two tokens as possible end tokens
- `vsg.token.case_generate_alternative.when_keyword`
- `vsg.token.case_generate_statement.end_keyword`

---

Should there be separate rules for the following?
- `elsif condition generate` -> `elsif condition generate begin`
- `else generate` -> `else generate begin`

### DIFF
--- a/docs/configuring_disabled_rules.rst
+++ b/docs/configuring_disabled_rules.rst
@@ -58,6 +58,7 @@ Rules Disabled by Default
 * `generic_600 <generic_rules.html#generic-600>`_
 * `generic_map_600 <generic_map_rules.html#generic-map-600>`_
 * `generic_map_601 <generic_map_rules.html#generic-map-601>`_
+* `if_generate_statement_001 <if_generate_statement_rules.html#if-generate-statement-001>`_
 * `instantiation_600 <instantiation_rules.html#instantiation-600>`_
 * `instantiation_601 <instantiation_rules.html#instantiation-601>`_
 * `loop_statement_006 <loop_statement_rules.html#loop-statement-006>`_
--- a/docs/configuring_disabled_rules.rst
+++ b/docs/configuring_disabled_rules.rst
@@ -47,6 +47,7 @@ Rules Disabled by Default
 * `context_ref_008 <context_ref_rules.html#context-ref-008>`_
 * `context_ref_009 <context_ref_rules.html#context-ref-009>`_
 * `file_100 <file_rules.html#file-100>`_
+* `for_generate_statement_001 <for_generate_statement_rules.html#for-generate-statement-001>`_
 * `function_600 <function_rules.html#function-600>`_
 * `function_601 <function_rules.html#function-601>`_
 * `generate_017 <generate_rules.html#generate-017>`_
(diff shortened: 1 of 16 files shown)


larshb:
The codeclimate-issues contradicts pre-commit styling and complaints about intentional duplication of the format/docstrings in the two rules.

JHertz5:
Hi @larshb.

> The codeclimate-issues contradicts pre-commit styling and complaints about intentional duplication of the format/docstrings in the two rules.

Don't worry about the codeclimate stuff, it throws errors for most PRs. As long as you've taken a look and solved issues where you can, you can ignore the remaining warnings.
Similar statement from the owner of the repository: https://github.com/jeremiah-c-leary/vhdl-style-guide/pull/1217#issuecomment-2336913532

jeremiah-c-leary:
Morning @larshb and @JHertz5 ,

> Don't worry about the codeclimate stuff, it throws errors for most PRs. 

I need to disable the codeclimate checks.  There were useful earlier in the development, but now they seem to be too noisy to be useful.

@larshb , I fear implementing this feature will be a little more difficult due to nesting generics.  Could you add some nested generates in your test cases?  Consider nesting `for_generate`, `if_generate`, and `case_generate`.  You could then use the same input for the three versions (for, if and case) of the rule.

Looking at the productions for the 
```mono
if_generate_statement ::=
generate_label :
  if [ alternative_label : ] condition generate
    generate_statement_body
  { elsif [ alternative_label : ] condition generate
    generate_statement_body }
  [ else [ alternative_label : ] generate
    generate_statement_body ]
  end generate [ generate_label ] ;
```
each `generate_statement_body` could have a `begin` keyword:
```mono
generate_statement_body ::=
  [ block_declarative_part
begin ]
  { concurrent_statement }
[ end [ alternative_label ] ; ]
```
So you could end up with the following code:
```vhdl
example_if:
if enable_fifo generate
    begin
elsif disable_fifo generate
    begin
else generate
    begin
end generate;
```
...or if you add in the `end` keywords:
```vhdl
example_if:
if enable_fifo generate
    begin
    end;
elsif disable_fifo generate
    begin
    end;
else generate
    begin
    end;
end generate;
```
Another issue to keep in mind is the `begin` keyword must separate the `block_declarative_part` and any `concurrent_statement`s.  Although this might take care of itself because the `begin` keyword is required if there is a `block_declarative_part`.

I think if you expand the test input we can better assess whether the rule as implemented provides the correct output in multiple scenarios.

btw, thanks for the pull request.

--Jeremy

jeremiah-c-leary:
Morning @larshb ,

Just wanted to ping you on this pull request to see if you still want to pursue it.

Thanks,

--Jeremy

jeremiah-c-leary:
Morning @larshb ,

I will assume you would not like to pursue this issue, so I will be closing it.  We can always revisit this in the future.

Regards,

--Jeremy

PR #1254, opened 2024-10-15
Issue#1251: Added rules to enforce case on parameters in functions and procedures

### LINKED ISSUE #1251: Case rules for formal parameters in functions and procedures
At the moment (as of v3.27.0), there doesn't appear to be any rule to enforce case for formal parameters in functions or procedures. After looking for such rules, I tested for their presence as follows:

Config:
Setting all case rules to "lower" and making sure that none are disabled.
```yaml
rule:
  group:
    case:
      case: "lower"
      disable: False
```
Code:
The parameter names of the function and procedure are in uppercase.
```
architecture rtl of test is

  procedure my_proc (
    PARAM1 : in integer;
    PARAM2 : out integer
  ) is
  begin

  end procedure my_proc;

  function my_func (
    PARAM1 : in integer;
    PARAM2 : out integer
  ) return integer is
  begin

  end function my_func;

begin

  my_proc (
           PARAM1 => MY_PARAM1,
           PARAM2 => MY_PARAM2
         );

  my_sig <= my_func (
                     PARAM1 => MY_PARAM1,
                     PARAM2 => MY_PARAM2
                   );

end architecture rtl;
```

Output
```
$ vsg -f test.vhd -c test.yml --fix
File:  test.vhd
Phase 7 of 7... Reporting
Total Rules Checked: 769
Total Violations:    0
  Error   :     0
  Warning :     0
```

I'd like to have rules that would enforce the case of formal parameters, e.g. changing the code above to:
```
architecture rtl of test is

  procedure my_proc (
    param1 : in integer;
    param2 : out integer
  ) is
  begin

  end procedure my_proc;

  function my_func (
    param1 : in integer;
    param2 : out integer
  ) return integer is
  begin

  end function my_func;

begin

  my_proc (
           param1 => MY_PARAM1,
           param2 => MY_PARAM2
         );

  my_sig <= my_func (
                     param1 => MY_PARAM1,
                     param2 => MY_PARAM2
                   );

end architecture rtl;
```

JHertz5:
Resolves #1251.

### DIFF
--- a/docs/function_rules.rst
+++ b/docs/function_rules.rst
@@ -583,6 +583,33 @@ This rule checks the function designator has proper case on the end function dec
 
    end function overflow;
 
+function_507
+############
+
+|phase_6| |error| |case| |case_name|
+
+This rule checks that the parameter names have proper case.
+
+|configuring_uppercase_and_lowercase_rules_link|
+
+**Violation**
+
+.. code-block:: vhdl
+
+  function my_func (
+    PARAM1 : in integer;
+    PaRaM2 : out integer
+  ) return integer;
+
+**Fix**
+
+.. code-block:: vhdl
+
+  function my_func (
+    param1 : in integer;
+    param2 : out integer
+  ) return integer;
+
 function_600
 ############
 
--- a/docs/ieee_rules.rst
+++ b/docs/ieee_rules.rst
@@ -1,9 +1,9 @@
 .. include:: includes.rst
 
-IEEE Rules
------------
+Ieee Rules
+----------
 
-IEEE_500
+ieee_500
 ########
 
 |phase_6| |error| |case| |case_keyword|
--- a/docs/procedure_call_rules.rst
+++ b/docs/procedure_call_rules.rst
@@ -293,3 +293,31 @@ This rule checks the **postponed** keyword has proper case.
 .. code-block:: vhdl
 
    postponed WR_EN(parameter)
+
+procedure_call_502
+##################
+
+|phase_6| |error| |case| |case_name|
+
+This rule checks that the parameter names have proper case.
+
+|configuring_uppercase_and_lowercase_rules_link|
+
+**Violation**
+
+.. code-block:: vhdl
+
+
+  my_proc (
+    PARAM1 => MY_PARAM1,
+    PaRaM2 => MY_PARAM2
+  );
+
+**Fix**
+
+.. code-block:: vhdl
+
+  my_proc (
+    param1 => MY_PARAM1,
+    param2 => MY_PARAM2
+  );
--- a/docs/procedure_rules.rst
+++ b/docs/procedure_rules.rst
@@ -771,3 +771,30 @@ This rule checks for consistent capitalization of procedure names.
      end process;
 
    end architecture rtl;
+
+procedure_508
+#############
+
+|phase_6| |error| |case| |case_name|
+
+This rule checks that the parameter names have proper case.
+
+|configuring_uppercase_and_lowercase_rules_link|
+
+**Violation**
+
+.. code-block:: vhdl
+
+  procedure my_proc (
+    PARAM1 : in integer;
+    PaRaM2 : out integer
+  );
+
+**Fix**
+
+.. code-block:: vhdl
+
+  procedure my_proc (
+    param1 : in integer;
+    param2 : out integer
+  );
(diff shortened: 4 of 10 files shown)


jeremiah-c-leary:
Afternoon @JHertz5 ,

> This also means that in cases where a set of nested subprograms share parameter names, there may be duplicated warnings, one from the inner subprogram and one from the outer subprogram. I'm not sure whether these issues can be tolerated or whether it is a dealbreaker. I can't imagine that there are many users who want the case for subprogram parameters to be different between a subprogram and a subprogram nested within that subprogram. Happy to keep working on it (maybe splitting consistent case out to a different issue) if desired though.


I found the consistent case rules to be difficult to implement given the current data structure VSG uses.  It is a lot of work trying to match the identifier with its use.

I think it would be a good idea to split the consistent case rule from the original request.  It would allow us to accept the rules for the arguments while the consistent case is being updated.

### Consistent case possible approach

One technique I have used before is to eliminate tokens from consideration.  Using the nested testcase as an example:
image

The arguments for `my_proc` on lines 5 through 9 do not apply from lines 15 through 97.  
The arguments for `my_func` on 16 through 20 only apply between 21 through 27.
The arguments for `my_proc` on lines 30 through 34 only apply between 36 and 58.

so if we can eliminate those lines from being considered by the analyze function, then they will not be checked and therefore will not be flagged as a possible problem.

One of the issues with the current data structure is hierarchy is not easily known.  In your testcase, knowing `my_func` is inside `my_proc` is not maintained, it must be determined because the data structure is just a list.  I have struggled with this for 6 years, but maybe the following idea could be a solution:

What if we record the production hierarchy in the token itself?  We know the hierarchy while going through the classify process, so we can maintain it there.  This would require a new attribute on the token class.  

For example, the `PARAM1` token on line 12 would have an attribute with the following list:
```python
[vsg.token.architecture_body, vsg.token.procedure_specification]
```
While the `param1` token on line 22 would have an attribute with the following list:
```python
[vsg.token.architecture.body, vsg.token.procedure_specification, vsg.token.function_specification]
```

So then we can grab all the tokens in the first procedure and filter them by asking "are you in my scope?".

This would be checked by taking the `vsg.token.procedure_specification` and seeing if it is the last `procedure_specification` or `function_specification` in the list.  If it is the last, then we know the token is in scope and need to check if it is the type of otken we are looking for.  If it is not the last, then we discard it and move on because we know it is not in our scope.

We could name the attribute `scope`.

What do think?

--Jeremy

JHertz5:
Hi @jeremiah-c-leary!

> I think it would be a good idea to split the consistent case rule from the original request. It would allow us to accept the rules for the arguments while the consistent case is being updated.

Yep I definitely agree, it makes perfect sense to split the PR up. I've raised #1310 for the consistent case rules, and I have cut those changes from this PR.

> What if we record the production hierarchy in the token itself? We know the hierarchy while going through the classify process, so we can maintain it there. This would require a new attribute on the token class.

This sounds like a great idea, I think that would make the problem much easier to solve,  and awareness of the heirarchy may come in handy in other situations too. I will aimm to have a go at this, although it might be a while before it is ready as I'm quite busy over the next couple of weeks.

>This would be checked by taking the vsg.token.procedure_specification and seeing if it is the last procedure_specification or function_specification in the list. If it is the last, then we know the token is in scope and need to check if it is the type of otken we are looking for. If it is not the last, then we discard it and move on because we know it is not in our scope.

This makes sense. One potential wrinkle to consider is that procedures and impure functions that are nested inside other subprograms would are able to use the parameters of the outer subprogram directly, e.g.
```vhdl
  procedure my_proc1 (
    i_param : in  integer;
    o_param : out integer
  ) is

    procedure my_proc2 (
      I_PARAM : in boolean
    ) is
    begin
      report to_string(I_PARAM) & to_string(O_PARAM);
    end procedure;

  begin
    O_PARAM := I_PARAM;
  end procedure;
```

In the inner subprogram, `o_param` is used and would need to be corrected, while `I_PARAM` within the inner subprogram overrules the `i_param` from the outer subprogram and wouldn't need to be corrected. So perhaps an alternative solution would be to scan through the procedures, keeping a stack of the various parameter names and popping items in and out if we find parameters with the same name as we traverse through the hierarchies.

I'll have a more in-depth think about it when I get around to working on it.

Many thanks,
Jukka

jeremiah-c-leary:
Morning @JHertz5 ,

> In the inner subprogram, o_param is used and would need to be corrected, while I_PARAM within the inner subprogram overrules the i_param from the outer subprogram and wouldn't need to be corrected. So perhaps an alternative solution would be to scan through the procedures, keeping a stack of the various parameter names and popping items in and out if we find parameters with the same name as we traverse through the hierarchies.

I did not have an appreciation for some of the nuances of VHDL until I started this project.

I will get this merged to master.

--Jeremy

JHertz5:
Thanks very much!

PR #1296, opened 2024-10-24
Issue#1287: Updated alignment rules to separate protected type bodies from other declarative regions

### LINKED ISSUE #1287: `architecture_029` aligns declarations within protected type declarative regions with declarations outside
**Environment**
VSG v3.27.0

**Describe the bug**
The rule `architecture_029` modifies the alignment of declarations within the declarative region of protected types as if they are a normal part of the architecture's declarative region. It's likely that other declarative regions in which protected types may be declared will suffer from similar problems.

**To Reproduce**
1. Create test file
```vhd
architecture rtl of test is

  type flag_pt is protected

    procedure set;

    procedure clear;

    impure function get return boolean;
  end protected;

  type       flag_pt is protected body
    variable flag : boolean;
  end protected body;

begin

end architecture rtl;
```
2. Run `vsg -f test.vhd --fix`
3. See error
```vhdl
architecture rtl of test is

  type flag_pt is protected

    procedure set;

    procedure clear;

    impure function get return boolean;
  end protected;
  --  vvvvvvv
  type       flag_pt is protected body
    variable flag : boolean;
  end protected body;

begin

end architecture rtl;
```

**Expected behavior**
```vhdl
architecture rtl of test is

  type flag_pt is protected

    procedure set;

    procedure clear;

    impure function get return boolean;
  end protected;

  type flag_pt is protected body
    variable flag : boolean;
  end protected body;

begin

end architecture rtl;
```

JHertz5:
Resolves #1287.
~~Built on top of PR #1293 so dependent on that being merged first.~~

In addition to excluding protected type bodies from various declarative region alignment rules, I've also added new rules for the declarative region of protected types.


JHertz5:
Marking as a draft until dependency has been merged.

JHertz5:
Dependency merged, ready for review.

jeremiah-c-leary:
Afternoon @JHertz5 ,

I understand the need for the PR to handle protected type differently.  For the protected type body tests, could you add some signals etc... outside of the protected type region.  Then we can ensure the test is only updating alignment within the protected type.

Thanks,

--Jeremy

JHertz5:
Hi @jeremiah-c-leary

That's a very good point, thanks for the feedback. I've updated the tests as requested.

Thanks,
Jukka

PR #1308, opened 2024-11-01
Issue#1291: Corrected corner case in multiline alignment rule

### LINKED ISSUE #1291: Missing one space in alignment of aggregate parts
The following file:
```
library ieee;
  use ieee.std_logic_1164.all;

entity b is
end entity b;

architecture a of b is

  signal s : MY_TYPE := (
                          a => '0',
                          ddddd => (others => '0'),
                          ffff => (others => '0')
                        );

begin

end architecture a;
```
With the configuration:
```yml
rule:
  signal_007:
    disable: yes
```
Produces:
```vhdl
library ieee;
  use ieee.std_logic_1164.all;

entity b is
end entity b;

architecture a of b is

  signal s : MY_TYPE := (
                          a => '0',
                         ddddd => (others => '0'),
                          ffff => (others => '0')
                        );

begin

end architecture a;
```
Notice the misalignment of `ddddd` by one space.

JHertz5:
Resolves #1291.
The fix is essentially
```diff
if len(lTemp) == 0:
+   if bIgnoreStartParen:
+       dExpectedIndent[iLine + 1] = lColumn[-1] + 1
+   else:
-   dExpectedIndent[iLine + 1] = lColumn[-1]
+       dExpectedIndent[iLine + 1] = lColumn[-1]
```
but I've made some additional changes to reduce code duplication (which I suspect is how the bug was introduced in the first place - there were three spots with identical code, and, when some change was made, only two spots were updated).


jeremiah-c-leary:
Afternoon @JHertz5 ,

> but I've made some additional changes to reduce code duplication (which I suspect is how the bug was introduced in the first place - there were three spots with identical code, and, when some change was made, only two spots were updated).

Code duplication, I should have taken care of that as part of my TDD process.  Thanks for fixing this.

PR #1311, opened 2024-11-10
Issue#1310: Add consistent case rules for paramter names in subprograms.

### LINKED ISSUE #1310: Add consistent case rules for parameter names in subprograms
I would like to be able to enforce consistent case of parameter names in functions and procedures, e.g. correcting
```vhdl
  procedure my_proc (
    i_param : in  integer;
    o_param : out integer
  ) is
  begin
    O_PARAM := I_PARAM;
  end procedure;
```
to
```vhdl
  procedure my_proc (
    i_param : in  integer;
    o_param : out integer
  ) is
  begin
    o_param := i_param;
  end procedure;
```

I would like new rules to enforce consistent case for the cases described above.

**Additional context**
This issue was split out from #1251, which is related to #1249. See [this comment](https://github.com/jeremiah-c-leary/vhdl-style-guide/pull/1254#issuecomment-2466469741) for details on how the solution for this issue may look.

JHertz5:
Resolves #1310.
~~Currently still a work in progress - see #1254 for details.~~
~~Dependent on #1254.~~

This is now updated and ready for review. I have updated the rules to respect heirarchy of nested subprograms. I think the change was less complicated than I was expecting. As always, any feedback, no matter how big or small, is welcome.

As part of these changes, some changes were made to `get_procedure_subprogram_body` and `get_function_subprogram_body`. The reason for these changes is that, before the change the functions are just checking for any `procedure`\ `function` keyword respectively in the TOI. As a result, `get_procedure_subprogram_body` would return a function if the function included a nested procedure, and vice-versa for `get_function_subprogram_body`. By checking only the first token, we limit the results to only the correct subprograms, and makes the code more efficient by
1. Checking fewer tokens.
2. Passing fewer tokens back to the client functions.


tests/procedure/rule_509_test_input.fixed.vhd
```diff
@@ -0,0 +1,220 @@
...
+    end procedure;
+
+    function my_func (
+      variable Param2 : out integer;
+      constant Param3 : in integer;
+      signal Param4   : out integer;
+      file Param5     : my_file
+    ) return integer is
+
+      constant con : integer := param1;
```
jeremiah-c-leary, line 169: Is the scope of `param1` correct?  It would seem is should be scoped to the function at line 162 instead of the procedure at line 139.

tests/procedure/rule_509_test_input.fixed.vhd
```diff
@@ -0,0 +1,220 @@
...
+      variable var : integer := param4;
+
+    begin
+
+      -- PARAM1
+      sig <= param1 + Param2;
+      var := param1 + Param2;
+      sig <= param1 when true else Param2;
+      var := param1 when true else Param2;
```
jeremiah-c-leary, line 178: I think `param1` should not have been changed since it is local to the function `my_func`.  I noticed the function does not include a `param1` argument though.

jeremiah-c-leary (changes requested):

tests/procedure/rule_509_test_input.fixed.vhd
JHertz5: That's a good point! This function needs to be declared `impure` to access `param1` from the parent subprogram, otherwise the code is invalid. I'll update the file.

tests/procedure/rule_509_test_input.fixed.vhd
JHertz5: `param1` is not declared within `my_func` (either as a parameter or as a variable/constant), so I don't think that it can be local to here. Since `param1` is not declared within `my_func`, I believe that the `param1` from the parameter list of the parent subprogram will be used. I've added the `impure` keyword to `my_func` to make sure that the code is valid.

JHertz5:
Thanks for your feedback @jeremiah-c-leary. I've actioned your comments and left some responses. The PR is ready for you to look at again.

tests/procedure/rule_509_test_input.fixed.vhd
jeremiah-c-leary: Evening @JHertz5 ,

Clearly I have never used an impure function.  I did read the LRM about pure and impure functions and agree with the case change as you have it shown.

It just seems you would want to always use a pure function.

--Jeremy

jeremiah-c-leary (approved):

tests/procedure/rule_509_test_input.fixed.vhd
JHertz5: Hi @jeremiah-c-leary

Yes, I've never come across a time where I've needed to use an impure function. I'd have thought that it would be undesirable for functions to have undeclared side-effects, but I suppose someone out there must have a use case for them to have included it in the language.

Thanks,
Jukka

PR #1351, opened 2024-12-13
use with FIPS

rpatel3001:
**Description**
hashlib.md5 needs `usedforsecurity=False` when used on FIPS enabled systems

using `new()` allows python < 3.9 to still work (when usedforsecurity was introduced)


jeremiah-c-leary:
Afternoon @rpatel3001 ,

Thanks for the pull request.  I never would have know the `usedforsecurity` was even a possible issue.

I will get this merged to master.

Thanks,

--Jeremy

PR #1356, opened 2024-12-14
Issue#1341: Added `delay_machanism` and `after` keyword case rules

### LINKED ISSUE #1341: Add case rules for delay_mechanism keywords
I'd like rules to enforce the case of delay_mechanism keywords, i.e. `transport`, `reject`, and `inertial`.
For example, these rules should be able to correct the code
```vhdl
architecture rtl of test is

begin

  test1 <= TRANSPORT 1 AFTER 10 ns;
  test1 <= REJECT 2 ns INERTIAL 0 AFTER 10 ns;

end architecture rtl;
```
to
```vhdl
architecture rtl of test is

begin

  test1 <= transport 1 after 10 ns;
  test1 <= reject 2 ns inertial 0 after 10 ns;

end architecture rtl;
```

JHertz5:
Resolves #1341.

### DIFF
--- a/docs/after_rules.rst
+++ b/docs/after_rules.rst
@@ -124,3 +124,24 @@ This rule checks the *after* keywords do not exist in the reset portion of a clo
        b <= c  after 1 ns;
      end if;
    end process clk_proc;
+
+after_500
+#########
+
+|phase_6| |error| |case| |case_keyword|
+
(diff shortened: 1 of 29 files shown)


JHertz5:
Marking as a draft - I still need to add a rule for the `after` keyword.

jeremiah-c-leary (approved):

PR #1441, opened 2025-04-19
fix SyntaxWarning for py3.13 compatibility

chenrui333:
**Description**
seeing some syntax warning when running with py3.13

```
/opt/homebrew/Cellar/vsg/3.31.0/libexec/lib/python3.13/site-packages/vsg/rules/whitespace/rule_011.py:16: SyntaxWarning: invalid escape sequence '\*'
  This rule checks for at least a single space before and after math operators +, -, /, * and \*\*.
```


jeremiah-c-leary:
Morning @chenrui333,

Thanks for  the PR.  I had the escapes to handle the double star not being bold.  It looks like the r in front of the double quote fixed that issue.

I will merge this to master.

--Jeremy

PR #1450, opened 2025-04-30
Add blank line rule in port_map

lgu-appear:
**Description**
Add a rule for detecting blank lines in the port map of an instantiation of a module.

**Screenshots**
See `tests\port_map\rule_201_test_input.vhd` and `tests\port_map\rule_201_test_input.fixed.vhd`.

### DIFF
--- a/docs/whitespace_rules.rst
+++ b/docs/whitespace_rules.rst
@@ -198,7 +198,7 @@ whitespace_011
 
 |phase_2| |error| |whitespace|
 
-This rule checks for at least a single space before and after math operators +, -, /, * and \*\*.
+This rule checks for at least a single space before and after math operators +, -, /, * and **.
 
 **Violation**
 


lgu-appear:
The test `test_board_cpu` fails because now the blank lines are removed from the port map. How should I proceed with that?

tests/port_map/test_rule_201.py
```diff
@@ -0,0 +1,49 @@
...
+lFile, eError = vhdlFile.utils.read_vhdlfile(os.path.join(sTestDir, "rule_201_test_input.vhd"))
+
+lExpected = []
+lExpected.append("")
+utils.read_file(os.path.join(sTestDir, "rule_201_test_input.fixed.vhd"), lExpected)
+
+
+class test_port_map_rule(unittest.TestCase):
+
+    maxDiff = None
```
benthie, line 21: Unused

benthie (changes requested):
Looks good as far as I can tell. I think the commits have to be squashed before this can be merged.

Could you also implement this rule as `generic_map_201`?

lgu-appear:
Good point, I can add the generic thing as you mention!

tests/port_map/test_rule_201.py
lgu-appear: I added this for pytest to display more log than limited

lgu-appear: Removed it for unicity over the repo

lgu-appear:
Hey! Thanks for reviewing. Can there be the possibility to include or not the blank line above the comments? In my case, I do not want any blank line, even above comments.

Loan

PR #1453, opened 2025-05-08
Fix file_rules globbing

### LINKED ISSUE #1425: file_rules globbing * or ** do not work
**Environment**
VHDL Style Guide (VSG) version: 3.30.0.post5+git.37d0bea7.dirty
Windows 10

**Describe the bug**
I have a file with port_011 violation at path `vivado/LED_blink/LED_blink.srcs/sources_1/new/top.vhd`.
I want to supress this violation for all files within `vivado` folder.
If I disable this rule for this specific file using file_rules in my `vsg.yaml` and run `vsg -c vsg.yaml -f vivado/LED_blink/LED_blink.srcs/sources_1/new/top.vhd` it works as expected and zero violations are reported:

```
rule:
    port_011:
        disable: false
file_rules:
    - vivado/LED_blink/LED_blink.srcs/sources_1/new/top.vhd:
        rule:
            port_011:
                disable: true
```

However, when I add any globbing pattern to the file path in the file_rules section, the violations are reported even though they should not be. The following paths all don't work (violation is reported and rule is not disabled):
```
- vivado/**/top.vhd:
- vivado/**/*.vhd:
- vivado/LED_blink/LED_blink.srcs/sources_1/new/*.vhd:
```

It seems to me, that globbing does not work for file_rules. Since the docs say that 

> The file_rules is exactly the same as file_list except that it will not add the file to the scan list.

and file_list accepts globbing, I expect that file_rules will also support globbing.

benthie:
Closes #1425

### DIFF
--- a/tests/comment/rule_010_test_input.fixed.vhd
+++ b/tests/comment/rule_010_test_input.fixed.vhd
@@ -16,22 +16,22 @@ context interfaces is
     -- use fpga.registers.all;
     use fpga.functions.all;
 
--- Comment last
+  -- Comment last
 
 end context interfaces;
 
 
 architecture RTL of FIFO is
 
--- failure
--- passed
--- failure
+  -- failure
+  -- passed
+  -- failure
 
 begin
 
--- failure
--- passed
--- failure
+  -- failure
+  -- passed
+  -- failure
 
 end architecture RTL;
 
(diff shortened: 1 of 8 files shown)


jeremiah-c-leary:
Afternoon @benthie and @JFMarten ,

It does appear that PR #1452 was used as a basis for this pull request.  I will work on removing those changes from this PR.

--Jeremy

PR #1455, opened 2025-05-09
Fix missing carriage return handling for rules moving token to previous line

### LINKED ISSUE #1444: Array definition misinterpreted as record definition if there's whitespace before semicolon
This file:
```vhdl
library ieee;
  use ieee.std_logic_1164.all;

entity b is
end entity b;

architecture a of b is

  type my_array0 is array (natural range 0 to 3) of std_logic;

  type my_array1 is array (natural range 0 to 3) of std_logic ;   -- white space before semicolon here

begin

end architecture a;
```
With the following rules:
```yaml
rule:
  whitespace_003:
    disable: yes
```
Produces the following output with vsg 3.31 `vsg -f b.vhd -c vsg.yml`:
```
Phase 1 of 7... Reporting
Total Rules Checked: 175
Total Violations:    1
  Error   :     1
  Warning :     0
-----------------------------+------------+------------+--------------------------------------
  Rule                       |  severity  |  line(s)   | Solution
-----------------------------+------------+------------+--------------------------------------
  record_type_definition_007 | Error      |         11 | Move **then** keyword to same line as std_logic
-----------------------------+------------+------------+--------------------------------------
```

benthie:
Closes #1444

### DIFF
--- a/vsg/rules/constant/rule_007.py
+++ b/vsg/rules/constant/rule_007.py
@@ -33,3 +33,4 @@ class rule_007(Rule):
     def __init__(self):
         super().__init__(lTokens)
         self.bRemoveTrailingWhitespace = False
+        self.solution = "Move := operator"
--- a/vsg/rules/entity/rule_024.py
+++ b/vsg/rules/entity/rule_024.py
@@ -33,3 +33,4 @@ class rule_024(Rule):
         super().__init__(lTokens)
         self.subphase = 2
         self.bRemoveTrailingWhitespace = False
+        self.solution = "Move entity simple name"
--- a/vsg/rules/entity/rule_025.py
+++ b/vsg/rules/entity/rule_025.py
@@ -33,3 +33,4 @@ class rule_025(Rule):
         super().__init__(lTokens)
         self.bInsertWhitespace = False
         self.subphase = 3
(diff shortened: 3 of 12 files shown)


jeremiah-c-leary:
Morning @benthie ,

I am having a hard time understanding the addition of the carriage return checking.  Could you elaborate on why it was added?

Thanks,

--Jeremy

benthie:
Hi @jeremiah-c-leary ,

to be honest, I can't recall completely. But if I remember correctly, the reason for the carriage return check had something to do with the fact that `record_type_definition_007` was triggered on a line like this

~~~
  type my_array0 is array (natural range 0 to 3) of std_logic ; -- whitespace before semicolon
~~~

although it should actually not get triggered at all since the semicolon is not on the next line. I think with the check I tried to prevent the rule from being triggered. But I just saw, that it still gets triggered on this PR branch if the semicolon is on the next line.

So I definitely have to write some tests. Maybe it then becomes clear again why I added this check an whether it is really necessary ;)

benthie:
Hi @jeremiah-c-leary , thanks for the prompt fix. I added two minor cosmetic fixes. The rest looks good to me :+1:

PR #1457, opened 2025-05-09
Extend rule concurrent_009 for manually broken lines 

### LINKED ISSUE #1456: Extend rule concurrent_009 for manually broken lines

When it comes to long lines, I sometimes want to break long concurrent assignments.

~~~vhdl
architecture a of b is

begin
  
  -- Breaking the long assignment like this ensures the length limit of 120 
  -- characters is met.
  my_very_long_concurrent_011_signal <=
      my_very_long_concurrent_003_signal(my_long_port_name_vector'length -1 downto 0) when select_value = '0' else
      my_very_long_concurrent_003_signal(3 downto 0);

end architecture a;
~~~

For this to work `concurrent_011`, which enforces no new line after assignment operator, must be disabled.

However, `concurrent_009` either has to be disabled as well, or `align_left` has to be set to `yes`. The problem is, that I would like to have the rule enabled and `align_left` set to `no`, so that the following format is nicely enforced:

~~~vhdl
architecture a of b is

begin

  my_concurrent_009_signal <= '0' when select_value = "1001" and foo = '0' else
                              '1';

end architecture a;
~~~

Which ultimately destroys the desired formatting of the first example. And the result again violates the line length limit of 120 characters.

~~~vhdl
architecture a of b is

begin

  -- Line exceeds line length limit!
  my_very_long_concurrent_011_signal <=
                                        my_very_long_concurrent_003_signal(my_long_port_name_vector'length -1 downto 0) when select_value = '0' else
                                        my_very_long_concurrent_003_signal(3 downto 0);

end architecture a;
~~~

~~~vhdl
architecture a of b is

begin
  
  -- This works
  my_very_long_concurrent_011_signal <=
      my_very_long_concurrent_003_signal(my_long_port_name_vector'length -1 downto 0) when select_value = '0' else
      my_very_long_concurrent_003_signal(3 downto 0);

  -- And this also works at the same time
  my_concurrent_009_signal <= '0' when select_value = "1001" and foo = '0' else
                              '1';

end architecture a;
~~~

With an additional value for `align_left`, e.g. `on_new_line_after_assign` one could enable this behavior. In the code we can check if the assignment is on a new line and if so, enforce left alignment.

benthie:
Closes #1456

### DIFF
--- a/docs/configuring_multiline_indent_rules.rst
+++ b/docs/configuring_multiline_indent_rules.rst
@@ -17,6 +17,9 @@ There are several options to these rules:
 .. |align_left__no| replace::
    :code:`no` = Lines will be aligned to assignment operator
 
+.. |align_left__on_new_line_after_assign| replace::
+   :code:`on_new_line_after_assign` = Lines will be aligned aligned left if there is a carriage return (alias “new line”) after the assignment operator.
+
 .. |align_paren| replace::
    :code:`align_paren`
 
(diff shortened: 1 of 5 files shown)


docs/configuring_multiline_indent_rules.rst
```diff
@@ -35,15 +38,17 @@ There are several options to these rules:
...
+| |align_left|  | |values| | |default_no|  | * |align_left__yes|                        |
+|               |          |               | * |align_left__no|                         |
+|               |          |               | * |align_left__on_new_line_after_assign|   |
++---------------+----------+---------------+--------------------------------------------+
+| |align_paren| | |values| | |default_yes| | * |align_paren__yes|                       |
+|               |          |               | * |align_paren__no|                        |
++---------------+----------+---------------+--------------------------------------------+
+
 
 This is an example of how to configure the option.
```
JHertz5, line 54: I suggest that you add an example below, like there are for the other settings, do demonstrate how the new setting behaves.

vsg/rules/multiline_conditional_alignment.py
JHertz5, line 1: There are no tests that cover the new settings. I think that you should add some tests to verify this functionality and to allow regression testing in the future.

benthie:
Hi @JHertz5 , I first wanted to wait for a feedback on the proposed feature before implementing the test cases. Maybe @jeremiah-c-leary could also comment on whether he thinks this feature makes sense implementing. If we agree that it is desired, I am happy to write the tests and update the documentation accordingly.

PR #1472, opened 2025-07-17
Issue#1471: UTF-8 check should handle output differences from command

### LINKED ISSUE #1471: Test utf_8
**Environment**

Version 3.33.0, Arch Linux

**Describe the bug**

The command

	file utf-8_encoded.vhd

returns in my system

	utf-8_encoded.vhd: Unicode text, UTF-8 text

This produces line

	self.assertTrue("UTF-8 Unicode text" in result.stdout)

to fail the test.

**To Reproduce**

Execute the full test set with pytest to get

	>       self.assertTrue("UTF-8 Unicode text" in result.stdout)
	E       AssertionError: False is not true

**Expected behavior**

`test_utf_8` should pass.

**Additional context**

Tested with `file-5.45` and `file-5.46`

linknum23:
Resolves #1471

### DIFF
--- a/tests/vsg/file_encodings/test_file_encodings.py
+++ b/tests/vsg/file_encodings/test_file_encodings.py
@@ -4,6 +4,7 @@ import pathlib
 import shutil
 import sys
 import unittest
+from unittest import mock
 from tempfile import TemporaryDirectory
 
 from vsg import __main__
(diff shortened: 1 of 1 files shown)


linknum23:
It looks like the static analyzer is pretty picky about how subprocess commands are used.  After a couple of tries I rewrote the encoding preservation checks just using pythons built-in file handling to avoid fighting with it.  I hope this makes sense. 

I added some internal validation of the test code, since so much changed.

This has a nice side effect of allowing these checks to happen on windows and github actions so I removed the `skipIf` checks and did a little more cleanup now that many parts were not longer used.

linknum23:
I left this as 2 commits for now so you can see the different solutions to this problem, they can easily be combined into 1 assuming the new encoding preservation checks make sense.

jeremiah-c-leary (approved):

PR #1476, opened 2025-07-23
Issue 1474: Add 2019 standard and its reserved keywords

### LINKED ISSUE #1474: Add VHDL 2019 Keywords for initial 2019 support
A simple way to start supporting VHDL 2019 is to add the new keywords introduced: private, view, vcpkg. 

This would disallow these keywords from being used as identifiers and provide a foundation to start building VHDL 2019 support on.

linknum23:
Resolves #1474

This adds the 2019 standard's reserved words, adds each of the reserved words to the relevant test and updates the reserved word documentation.

Should we include the 2019 standard in the default all? I'm on the fence.


vsg/rules/reserved/rule_001.py
```diff
@@ -538,6 +538,125 @@ def uniquify_standards(dMap):
     "xnor",
     "xor",
 ]
+# https://www.0x04.net/~mwk/vstd/ieee-1076-2019.pdf, p. 266-267, 15.10: Resserved Words
```
JHertz5, line 541:
```suggestion
# https://www.0x04.net/~mwk/vstd/ieee-1076-2019.pdf, p. 266-267, 15.10: Reserved Words
```
Spelling mistake.

JHertz5 (changes requested):

PR #1565, opened 2026-06-15
Issue#1564: Fix block_201 false positive on leading comments when block "is" is omitted

### LINKED ISSUE #1564: `block_201` raises an error on comment lines when `is` is missing
**Environment**
git hash a895557eee9ce0fb9355828758f217c11563aa47 (HEAD of the master branch at time of writing)

**Describe the bug**
`block_201` is flagging errors on code that isn't even within a block statement when the `is` of the block statement (which is optional) is missing and the file starts with comments. If the `is` is present, or if the comments are removed, the warning goes away. If `--fix` is used, VSG "fixes" the issue by putting a blank line between the comment lines at the top.

**To Reproduce**
1. RTL file: test.vhd
```vhdl
--
--

architecture rtl of test is
begin
  my_label : block

  begin
  end block;
end architecture;
```
2. Config file: test.yml
```yaml
rule:
    global:
        disable: true
    block_201:
        disable: false
```
3. Run `$ vsg -f test.vhd -c test.yml`
```
File:  test.vhd
Phase 3 of 7... Reporting
Total Rules Checked: 1
Total Violations:    1
  Error   :     1
  Warning :     0
------------+------------+------------+--------------------------------------
  Rule      |  severity  |  line(s)   | Solution
------------+------------+------------+--------------------------------------
  block_201 | Error      |          1 | Insert blank line below
------------+------------+------------+--------------------------------------
```

**Expected behavior**
There should be no error in this case, `block_201` should be satisfied by this code.

gaoflow:
## Description

Fixes #1564.

`block_201` was reporting a violation against the first line of a file (a leading comment), and `--fix` inserted a spurious blank line between the leading comments, whenever a `block` statement omitted its optional `is` keyword and the file contained no other `is`/guard tokens.

The cause is in `get_indexes_of_tokens_between` (`vsg/vhdlFile/extract/utils.py`). When a start token (here the `block` keyword) has no following end token, the function appended `iStartIndex`, the `enumerate` loop counter, instead of `iStart`, the actual token index. With a single un-`is`-ed block the counter is `0`, so the rule was evaluated against token index `0`, the file's leading comment. The fix appends the real token index so the rule is evaluated against the block line.

## Testing

- Added `tests/block/rule_201_test_input.issue_1564.vhd` and `test_rule_201_does_not_flag_leading_comments`, which fails before the fix (violation reported on line 1) and passes after it.
- Verified the original report: the bogus `block_201` error is gone and `--fix` leaves the leading comments untouched, while a genuine missing-blank-line below a block is still flagged on the correct line.
- The existing `block_201` tests and the other rules that use this helper (`block_003`, `context_010`/`011`, `loop_statement_004`) pass unchanged.

I use AI tooling to help with this work, under my direction and review; the analysis, fix, and tests were checked by me.


tests/block/rule_201_test_input.block_without_is.vhd
JHertz5, line 1: The naming of this file is not consistent with the rest of the codebase. It would be preferable if you were able to incorporate this test case into the existing `rule_201_test_input.vhd` file.

tests/block/test_rule_201.py
```diff
@@ -71,3 +71,22 @@ def test_fix_rule_201_no_blank_line(self):
 
         oRule.analyze(self.oFile)
         self.assertEqual(oRule.violations, [])
+
+
+lFile_issue_1564, eError_issue_1564 = vhdlFile.utils.read_vhdlfile(os.path.join(sTestDir, "rule_201_test_input.issue_1564.vhd"))
+
+
+class test_rule_issue_1564(unittest.TestCase):
```
JHertz5: Creating a separate class for another test is not consistent with the rest of the codebase. It would be preferable to incorporate this test into the `test_rule` class rather than creating a separate class. This is also useful because it allows us to test this case against each of the style options.

JHertz5 (changes requested):
@gaoflow Thanks very much for opening this PR. I can confirm that this change does fix the problem on my end as well. The only issue that I have with the PR is that the creation of a new test file and test class is not consistent with the existing testing architecture.

binggao1230:
Thanks for the review! I've moved the test into the `test_rule` class and it now checks both the `require_blank_line` and `no_blank_line` style options (under the bug it misreported `[1]` and `[]` respectively; fixed it's `[]` and `[9]`).

On the input file: this particular case can't be folded into `rule_201_test_input.vhd`. The bug only triggers when the file contains *no* block `is`/guard tokens at all (that's what leaves `lEndIndexes` empty and hits the buggy branch in `get_indexes_of_tokens_between`). The shared input file is full of `is`/`(guard)` blocks, so adding the scenario there masks the bug entirely. I've instead followed the existing multi-input convention (e.g. `block_comment/rule_001_test_input.align_left.vhd`) and renamed it to `rule_201_test_input.block_without_is.vhd`. Happy to adjust further if you'd prefer a different name.

binggao1230:
On folding this into `rule_201_test_input.vhd`: I tried that first, but the bug only manifests when the file contains *no* `is`/guard end-tokens at all — that's the `except IndexError` branch in `get_indexes_of_tokens_between` that this PR fixes. `rule_201_test_input.vhd` is full of `is`/guard tokens, so the case can't reproduce there. I verified this: reverting the one-line fix fails *only* `test_rule_201_block_without_is_keyword`, while all the shared-file tests stay green — a folded test would pass with or without the fix and wouldn't guard the regression.

For consistency I followed the existing `tests/block_comment/test_rule_001.py` pattern, which loads a scenario input (`rule_001_test_input.align_left.vhd`) alongside the main file and exercises it from a dedicated method inside the single `test_rule` class. That's also where I moved this test — it's no longer a separate class and now checks both `require_blank_line` and `no_blank_line`. Happy to rename the input file if you'd prefer a qualifier other than `block_without_is`.

JHertz5 (approved):
Thanks @gaoflow! Apologies for the delay, work has been hectic. The changes look good to me.

